PR #27, opened 2024-09-25
fix: remove jdk version info

robinyeeh:
1. update class file header.
2. remove jdk info from class description


spring-ai-alibaba-core/src/main/java/com/alibaba/cloud/ai/advisor/DocumentRetrievalAdvisor.java
```diff
@@ -1,8 +1,19 @@
 /*
- * All rights Reserved, Designed By Alibaba Group Inc.
- * Copyright: Copyright(C) 1999-2024
- * Company  : Alibaba Group Inc.
+ * Copyright 2023-2024 the original author or authors.
```
yuluo-yx, line 2: The code first started in 2024, maybe it is more appropriate to change it to 2024

spring-ai-alibaba-core/src/main/java/com/alibaba/cloud/ai/advisor/DocumentRetrievalAdvisor.java
```diff
@@ -21,11 +32,9 @@
 /**
  * Title Document retrieval advisor.<br>
  * Description Document retrieval advisor.<br>
- * Created at 2024-08-27 10:47
  *
  * @author yuanci.ytb
- * @version 1.0.0
- * @since jdk8
+ * @since 2024/8/16 11:29
```
yuluo-yx, line 37: It is more common to use version numbers instead of dates, `@since 1.0.0-Snapshot`

PR #59, opened 2024-10-12
feat:add ollama example

wblu214:
Help users quickly use spring ai alibaba to call other large models, including locally deployed ones such as ollama


yuluo-yx (approved):
LGTM

PR #63, opened 2024-10-16
feat: add bailian multimodel support

robinyeeh:
1. add bailian multimodel support (bailian vl models).
2. able to understand image and video.

None

### Describe how to verify it
Verified with test cases.


spring-ai-alibaba-core/src/test/resources/multimodel/dog_and_girl.jpeg
yuluo-yx, line 1: This file must exist. I see that the url address is used in the test class.

yuluo-yx (approved):
LGTM

PR #65, opened 2024-10-20
fix AdvisedRequest import path

conghuhu:
Due to changes in the path of AdvisedRequest in spring ai core, the code in the example needs to be modified

no

### Describe how to verify it
image


chickenlj (approved):
LGTM.

PR #70, opened 2024-10-24
feat:add user agent

robinyeeh:
1. Update dashscope api user agent.
2. Fix metadata of ChatResponse loss issue in advisor.

### Describe how to verify it
Verified with test cases.


yuluo-yx (approved):
LGTM

PR #75, opened 2024-10-28
 update: analyticdb+postgresql vectore

sincerity-being:


spring-ai-alibaba-core/src/main/java/com/alibaba/cloud/ai/dashscope/rag/AnalyticdbVector.java
```diff
@@ -0,0 +1,270 @@
...
+				throw new RuntimeException(
+						"Failed to create collection " + this.collectionName + ": " + e.getMessage());
+			}
+		}
+	}
+
+	@Override
+	public void add(List<Document> texts) {
+		try {
+			initialize();
```
chickenlj, line 152: How about using `afterPropertiesSet()` to do initialization. It guarantees only execute once and before any other method calls.

sincerity-being:
once again update

PR #96, opened 2024-11-07
feat: add DashScopeChatModel call and stream observability

Cirilla-zmh:
Add DashScopeChatModel call and stream observability so that we can trace and generate metrics for `DashScopeChatModel`.

### DIFF
--- a/spring-ai-alibaba-core/pom.xml
+++ b/spring-ai-alibaba-core/pom.xml
@@ -105,6 +105,12 @@
 			<artifactId>reactor-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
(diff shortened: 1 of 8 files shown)


robinyeeh:
LGTM

chickenlj (approved):
LGTM.

liugddx:
Why implement it here instead of contribute to spring ai?

Cirilla-zmh:
> Why implement it here instead of contribute to spring ai?

Instrumentation should be integrated with the instrumented components, so I can only provide an implementation here.
If you are wondering why the DashScope ChatModel is not implemented in Spring AI, perhaps other maintainers can provide more details.
cc @chickenlj @robinyeeh

PR #112, opened 2024-11-19
Baidu search plugin

CZJCC:


community/baidu-search-plugin-starter/src/main/java/com/alibaba/cloud/plugin/BaiduSearchPluginConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+ * @since : 2024-11-18
+ **/
+
+@Configuration
+@ConditionalOnClass(BaiduSearchService.class)
+public class BaiduSearchPluginConfiguration {
+
+	@Bean
+	@ConditionalOnMissingBean
+	@Description("Use baidu search engine to query for the latest news.") // function
```
chickenlj, line 41: I guess here we misused `jdk.jfr.Description`, it should be `org.springframework.context.annotation.Description`.

community/baidu-search-plugin-starter/src/main/java/com/alibaba/cloud/plugin/BaiduSearchPluginConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+																			// description
+	public BaiduSearchService baiduSearchService() {
+		return new BaiduSearchService();
```
chickenlj, line 44: Is it necessary to give users a method to set the behavior of BaiduSearch plugin? For example, by defining a  `BaiduSearchProperties` bean. 

It depends on the design of BaiduSearch. If it requires a token/key from user, then we must provide a way for users to set it. If it has some widely used parameters that will affect how it runs and the final result, we'd better make them configurable.

chickenlj (changes requested):

community/baidu-search-plugin-starter/src/main/java/com/alibaba/cloud/plugin/BaiduSearchPluginConfiguration.java
chickenlj:
```java
public BaiduSearchService baiduSearchService(BaiduSearchProperties properties) {
		return new BaiduSearchService(properties);
}
```

chickenlj (approved):
LGTM.

PR #125, opened 2024-11-21
realize chatModel

jianyi-gronk:

- realize chatModel
- fix ReferenceError, window is not defined


yuluo-yx (approved):
LGTM

PR #134, opened 2024-11-26
Feat: Intergration Weather PluginTool

Yeaury:

add a plugin to get the weather function

It is used to obtain the current weather and forecast weather information of the corresponding city

#98


chickenlj (commented):
LGTM.

Issue #199, opened 2024-12-13
Calling Function always throws Exception:No function callback found for function name


Mrs-Snow:
Spring-ai-alibaba
Calling Function always throws Exception:No function callback found for function name, because the function name's value is always "" which I found by debug the code of DashScopeChatModel.

I need the progress do not throw the exception instead of telling user other info or abandon calling function.
I also want to known how can I prevent from this exception?

Mrs-Snow:
and I also found,when the function has param 'chatId', the param's value will be modified to 1234567890

hongshuo-wang:
I encountered the same exception while using the stream() function; however, the problem does not occur when I use call().

Mrs-Snow:
> I encountered the same exception while using the stream() function; however, the problem does not occur when I use call().

Modify the class DashScopeAiStreamFunctionCallingHelper(line142) to 'String name = (current.name() != null && current.name().length() > 0 ? current.name() : previous.name());' The problem was fixed!

Mrs-Snow:
> I encountered the same exception while using the stream() function; however, the problem does not occur when I use call().

using call will avoid method merge，so it does not occur。

Issue #215, opened 2024-12-17
streaming output with function call throws exception: No function callback found for function name: 


hongshuo-wang:
Function Calling

streaming output with function call throws exception: No function callback found for function name: ,the name of function in toolCalls parameter is empty

you can reproduce this bug with official tutorial sample code, in 1.0.0-M3.2 or 1.0.0-M2.1 version which I've tried.

采用流式输出函数调用结果，会抛异常No function callback found for function name: ，debug发现找不到toolCalls变量里面的function的name，并且之前有一个测试过是正常的代码，现在调用也不正常了，怀疑是百炼平台的流式api接口出了点问题。

LittleDrip:
我也出现了这个问题，openAI的就可以用的，希望尽快解决

hongshuo-wang:
> 我也出现了这个问题，openAI的就可以用的，希望尽快解决

https://github.com/alibaba/spring-ai-alibaba/issues/199#issuecomment-2550006265
This solution works for now, but it's not very convenient as it requires modifying the source code. I hope that the official next version can include this fix as soon as possible. It would be great if it could be urgently addressed since it has a significant impact.

目前采用这个方案解决了，但是要改源码，还是不太方便，希望官方下个版本尽快上，能紧急修复下吧，影响还挺大的

lovesupertiger:
遇到同样的问题,该怎么解决?

lovesupertiger:
谢谢

lovesupertiger:
@PolarishT @Cirilla-zmh 谢谢

Cirilla-zmh:
@hongshuo-wang @lovesupertiger @LittleDrip 
Hey, seems to be a lot of similar questions being asked at the same time. I have fixed this issue based on @PolarishT's PR and hope it could work well. (I could work as expected on my machine now)

LittleDrip:
> @hongshuo-wang @lovesupertiger @LittleDrip Hey, seems to be a lot of similar questions being asked at the same time. I have fixed this issue based on @PolarishT's PR and hope it could work well. (I could work as expected on my machine now)

Why do I still have the above error? Is it because the Taobao image source I am using has not been updated？

Cirilla-zmh:
> Why do I still have the above error? Is it because the Taobao image source I am using has not been updated？

This fix has just been merged into the code repository and is not yet included in latest release. So if you want to verify this fix, you will need to package it locally first.

[...]

PR #217, opened 2024-12-18
[hotfix]: fix stream function calling exception

PolarishT:

1. support function calling stream output

Fixes: #199
#215 

### Describe how you did it

```java
String name = (current.name() != null && !current.name().isEmpty() ? current.name() : previous.name());
```


chickenlj (approved):
LGTM.

Cirilla-zmh:
@PolarishT 
Hey, I have noticed similar issue and thanks for your contribution. However, this PR seems that the exception couldn't be completely avoid.

I made some additional modifications based on your changes. According to the IT results, it is now working well!

https://github.com/alibaba/spring-ai-alibaba/pull/220

PR #220, opened 2024-12-19
fix IllegalStateException caused by the inability to parse the id and name of the function call

### LINKED ISSUE #199: Calling Function always throws Exception:No function callback found for function name

Cirilla-zmh:
fix #199 #215 
related to #217 

ITs of dashscope chat model function calling are also fixed.


chickenlj (approved):
LGTM.

Cirilla-zmh:
There are two commits and the second one is just adjusting the code format so you can ignore it. ;)

Dawn123F:
这个我跑官方demo也会报这个错误

PR #349, opened 2025-01-04
feat(studio): add chat memory func &  fix trace in chatClient

conghuhu:
#### chat memory
image

#### trace
image

#### clear history context
image


conghuhu:
@chickenlj PTAL

PR #358, opened 2025-01-08
code Format

heliang666s:
I used the command mvn spring-javaformat:apply to format the code in this branch. Otherwise, every time the project is packaged or compiled, errors would occur, which significantly impacts the efficiency of developers.


chickenlj:
Please submit a new one based on and targeted to the main branch.

PR #405, opened 2025-02-18
rm AutoConfiguration in core

stillmoon:

fix bug in maven dependency. When the spring-ai-alibaba-core package is introduced separately and Dashscopechatmodel is manually initialized, the SPRINGBOOT project fails to start. Because of the autoconfiguration of spring-ai-alibaba-core. Imports declares objects that don't belong to this package. In fact, this declaration has been moved to spring-ai-alibaba-autofigure, and the declaration in the core should be removed.

### Describe how you did it

rm AutoConfiguration.imports in spring-ai-alibaba-core


yuluo-yx (approved):
LGTM, thanks

chickenlj:
I think we need to move the configuration from `core` to `autoconfigure`

PR #413, opened 2025-02-21
[reactor]: reactor functioncalling 2 toolcalling

PolarishT:
1. current spring ai marks the function-calling deprecated.


PolarishT:
cc @chickenlj

PR #423, opened 2025-02-25
replacing getContent() API to getText()

kaori-seasons:
After upgrading spring-ai-alibaba, some old APIs of Document Reader API cannot be used and need to be replaced with new APIs.


yuluo-yx (approved):
LGTM

yuluo-yx:
ptal @PolarishT

PR #449, opened 2025-03-07
Unify the names of the keys and add IT use cases for chat.

brianxiadong:
Unify the names of the keys and add IT use cases for chat.


yuluo-yx (approved):
lgtm

PR #465, opened 2025-03-12
reduce redundant getProperty calls to improve performance

qnnn:
From the JMH test results, getProperty usually incurs some overhead.


yuluo-yx (approved):
lgtm, thanks

PR #641, opened 2025-04-13
fix: lark suite config and receive_id_type issue

kone-net:
1. Lark suite function call config is not right. 
2. Lark message need to add one more parameter receive_id_type.

### Describe how you did it
add receive_id_typed parameter and change config

### Describe how to verify it
I test it on my device.

image
image


yuluo-yx:
单测似乎失败了，修一下？

kone-net:
> 单测似乎失败了，修一下？

看着不是修改的代码引起的，合并了最新main分支代码，可以再看看

PR #652, opened 2025-04-15
fix java.lang.NullPointerException: Cannot invoke "com.alibaba.cloud.…

chiefleo:
…ai.dashscope.api.DashScopeApi$ChatCompletionOutput.choices()" because the return value of "com.alibaba.cloud.ai.dashscope.api.DashScopeApi$ChatCompletionChunk.output()" is null


yuluo-yx (approved):
LGTM

PR #685, opened 2025-04-21
Support mcp server auto register to nacos

Sunrisea:
Support mcp server auto register to nacos


chickenlj (approved):
LGTM.

PR #696, opened 2025-04-21
disable baidu test

zhangshenghang:

Because anti crawling is temporarily unsuccessful, disable this method first and enable it after it is fixed.


yuluo-yx (approved):
lgtm

PR #730, opened 2025-04-26
Feat: Add HttpNode to Graph Module

Yeaury:

This PR introduces a new HttpNode into the graph execution engine, enabling fully-configurable HTTP requests with support for dynamic variables, response parsing (including file extraction), and automatic retry on transient errors. Comprehensive unit tests using MockWebServer ensure correct behavior in both JSON/text and file-attachment scenarios.

NONE

#### Initialization Parameters

Configure HTTP method (GET, POST, PUT, DELETE, etc.)

Set target URL, headers, and query parameters

Define request body (raw text, JSON, form-urlencoded, multipart form-data, binary)

Support Basic and Bearer authentication

#### Dynamic Template Variables

Inject variables into URL, headers, query params, and body via ${variableName} syntax

Values are resolved at runtime from the OverAllState context

#### Response Handling & File Extraction

Parse HTTP status code, headers, and response body

Detect file-attachment responses via Content-Disposition or binary content types

Automatically cache file bytes in in-memory storage and expose a File DTO for downstream nodes

#### Error Retry Policy

Built-in retry mechanism on transient failures (network timeouts, 5xx responses)

Default: 3 retries with a 5 000 ms max backoff interval

Configurable per-node via RetryConfig

### Special notes for reviews


chickenlj:
Well done.

chickenlj (approved):

PR #734, opened 2025-04-27
feat: add ListOperatorNode to graph module

VLSMB:

The model, when accepting elements of a list such as numbers, strings, or file objects, may need to filter based on specific conditions or sort the list to pass the data to the next Node for processing, thus requiring a ListOperatorNode.

NONE

### Describe how you did it

A generic class `ListOperatorNode` is defined to generate node objects capable of handling different list elements. For consistency, it is specified that `T` must be a subclass of the `ListOperatorNode.ListElement` interface. Predefined classes `NumberElement`, `StringElement`, and `FileElement` are provided to handle lists containing numeric values (a mix of Integer and Double), strings, and file objects respectively. The `FileElement` corresponds to a JSON-formatted file object. These three classes include predefined methods that can be instantiated as Predicate and Comparator, such as checking if a numeric element is an integer, excluding files with specific extensions, sorting by file size, etc.

The apply method of the Node retrieves a list JSON string from the `OverAllState` object (either from the previous node or user input), then performs filtering, sorting, and element count limitation based on the Predicate and Comparator combinations passed during the instantiation of the `ListOperatorNode`. Finally, the processed list is converted back into a JSON string for output.

### Describe how to verify it

A sample project was written to verify that the node's output meets expectations.


disaster1-tesk:
It is hoped to submit a unit test class to help understand the usage of node

VLSMB:
> It is hoped to submit a unit test class to help understand the usage of node

okay, I'll make it.

PR #783, opened 2025-05-07
improve

zhangshenghang:

We need to set the default thread ID, otherwise it will all be saved together.

Before:
928c48c1e047266f76f120aa139d07df

After:
image


yuluo-yx (approved):
lgtm

PR #807, opened 2025-05-09
[feat] add param doc, change default enable_thinking value

Aias00:
… value

- For #786 

add param doc, change default enable_thinking value

### Describe how you did it
add param doc, change default enable_thinking value


yuluo-yx (approved):
LGTM

PR #813, opened 2025-05-10
ci: fix newline check ci

yuluo-yx:
- For #786


godofcodeeyes:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #843, opened 2025-05-13
feat(embedding): embedding model compatibale with M8 

jackxu2011:

- For #786 

embedding model compatibale with M8


yuluo-yx (approved):
LGTM

PR #887, opened 2025-05-19
Human node

yangzehan:
Add humanFeedbackMessage field to HuamNode


disaster1-tesk:
I hope there will be individual tests to help everyone understand

PR #902, opened 2025-05-20
[fix] fix elasticsearch memory pom

Aias00:
 fix elasticsearch memory pom


yuluo-yx (approved):
LGTM

PR #910, opened 2025-05-20
modify dependencies optional scope

Aias00:

modify dependencies optional scope for mcp nacos


yuluo-yx (approved):
LGTM

PR #942, opened 2025-05-23
refactor: use OkHttp in uploadFile to preserve empty Content-Type as …

joerulau:
…required by upload lease API

Refactored the uploadFile method. In the previous step, when applying for a document upload lease, the HTTP request parameters returned for uploading the document may have a Content-Type that is null. However, the server requires that the empty Content-Type be preserved during the upload. When using RestTemplate, if Content-Type is null, it gets removed automatically, causing the request to be incorrect. Therefore, the implementation has been updated to use OkHttp for the upload, ensuring that the Content-Type field, even if empty, remains in the request.


joerulau:
这个修改为啥没有合并到主分支呀？

PR #1005, opened 2025-05-28
fix: fix graph Concurrent call exception

jackxu2011:

1. 修复 graph 并发调用出错bug
2. DashScopeAiStreamFunctionCallingHelper.java重构部分代码


jackxu2011:
cherry-pick to #1015

PR #1009, opened 2025-05-28
feat: add source URL in Baidu search results

DomKing:
current search results only contains title and abstractText, the source url is also useful in special condition.

NONE

### Describe how you did it
extract sourceUrl from div content

### Describe how to verify it
write test code and check it

image


DomKing:
Checks failed at `spring-ai-alibaba-graph-core module`, but not `spring-ai-alibaba-starter-tool-calling-baidusearch`

chickenlj (approved):

PR #1013, opened 2025-05-28
fix: Exception in ReporterNode

VLSMB:

fix: The client in reporterNode is missing the prompt configuration, causing an HTTP 400 error.


GTyingzi:
LGTM

PR #1024, opened 2025-05-29
Fix resume 2

yangzehan:
The CompiledGraph.resome method requires updating graph data, while the evaluateAction method retrieves the state from the original graph


disaster1-tesk:
这个明天graph会做优化，pr可以关闭哈，感谢提pr

disaster1-tesk:
是已经发现的问题

PR #1058, opened 2025-06-01
Feat: Run Python Code in Docker for CoderNode of Deepresearch

VLSMB:

The previous version of the code directly executed Python code from the AI model locally without additional code review or resource restrictions, which could lead to infinite loops causing the project to freeze or the AI model delivering destructive code. To ensure security, the Python code should now run within Docker containers. Additionally, users can configure restrictions for the Python code, such as network access and memory resources.


GTyingzi:
LGTM

PR #1078, opened 2025-06-03
fix(mcp): fix the bug of different version of mcp servers register to same service in Nacos

Sunrisea:

Fix the bug of different version of mcp servers will register to same service in Nacos


Aias00 (approved):
lgtm

PR #1080, opened 2025-06-04
fix(mcp): fix mcp dep and config metadata

Aias00:

fix(mcp): fix mcp dep and memory dep and config metadata


PolarishT (dismissed):
LGTM.

PR #1084, opened 2025-06-04
fix(jmanus):fix the conversation memeory

skyXJJ:

fix the conversation memeory.
When no conversationId is passed from the frontend, provide a default value to ensure context continuity within a single page.


skyXJJ:
drop

PR #1166, opened 2025-06-11
feat(graph): add plugin node

Aias00:
1. add plugin node
2. add weather plugin
3. add nacos plugin


chickenlj:
I doubt whether `PluginNode` is necessary to act as a proxy Node for `GraphPlugin`. There's no real logic in `PluginNode`, the only thing it does is to get the param from state and write the output of `GraphPlugin` to state. So that means implementing `NodeAction` or `GraphPlugin` doesn't have any differences for developers.

Aias00:
> I doubt whether PluginNode is necessary to act as a proxy Node for GraphPlugin. There's no real logic in PluginNode, the only thing it does is to get the param from state and write the output of GraphPlugin to state. So that means implementing NodeAction or GraphPlugin doesn't have any differences for developers.

Got it. I'll close this pr right away. I realize I need to have a more in - depth understanding of NodeAction. I'll take the time to study it carefully and explore more possible directions. Thanks for your feedback!

PR #1201, opened 2025-06-13
fix(nl2sql): remove repeatedly lomobk.

yangzehan:
Lombok has repeatedly declared dependencies


Aias00:
hi, can u just remove this lombok?

PR #1217, opened 2025-06-14
feat(deepresearch): add crawler to researcher node

VLSMB:

Add Jina Reader API to Researcher and Background Node. And this plugin can be chosen to enable or not use.

Jina Crawler plugin is from PR #1189.


GTyingzi (approved):
LGTM

PR #1277, opened 2025-06-18
fix(core): update JSON serialization to use UTF-8 encoding

zhangshenghang:

update JSON serialization to use UTF-8 encoding


Aias00 (approved):
LGTM

PR #1288, opened 2025-06-19
feat(nl2sql): implementation of NL2SQL Function Based on Graph

zhangshenghang:

implementation of NL2SQL Function Based on Graph

### DIFF
--- a/spring-ai-alibaba-nl2sql/chat/README.md
+++ b/spring-ai-alibaba-nl2sql/chat/README.md
@@ -156,6 +156,320 @@ chatbi:
 
 ---
 
+## NL2SQl Graph
+
+### 设计
+
+```plantuml
+@startuml
+skinparam usecaseFontSize 14
+skinparam usecaseStereotypeFontSize 12
+skinparam hexagonFontSize 14
+skinparam hexagonStereotypeFontSize 12
+title "workflow graph"
+footer
+
+powered by spring-ai-alibaba
+end footer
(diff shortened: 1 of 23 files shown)


spring-ai-alibaba-nl2sql/chat/README.md
```diff
@@ -156,6 +156,320 @@ chatbi:
 
 ---
 
+## NL2SQl Graph
+
+### 设计
```
yuluo-yx, line 161: Github md 似乎不支持这种格式的渲染？

spring-ai-alibaba-nl2sql/chat/src/main/java/com/alibaba/cloud/ai/config/Nl2sqlConfiguration.java
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * @author zhangshenghang
+ */
+@Configuration
+public class Nl2sqlConfiguration {
+
+	private static final Logger logger = LoggerFactory.getLogger(Nl2sqlConfiguration.class);
+
+	@Autowired
```
yuluo-yx, line 59: IMO, in the project, we can use constructor injection instead of annotation injection to keep it consistent with the core module. Note that this is just a tips, not a mandatory requirement.

yuluo-yx (approved):

PR #1305, opened 2025-06-20
feat(deepresearch): adjust mcp

GTyingzi:
代码重构，主要是分配mcp到指定node节点

### Describe how to verify it
image


zxuexingzhijie (approved):

GTyingzi:
> hi @GTyingzi. IMO, in the project, we can use constructor injection instead of annotation injection to keep it consistent with the core module. Note that this is just a tips, not a mandatory requirement.

ok，这块提醒收到了

PR #1310, opened 2025-06-21
fix(core): add `spring-webflux` dependency

YunKuiLu:
fix org/springframework/web/reactive/function/client/WebClient$Builder Class Not Found.

The issue occurs because `DashScopeApi` has a mandatory dependency on `org.springframework.web.reactive.function.client.WebClient`, but the `spring-webflux` dependency is missing in the `pom.xml`, which leads to an exception being thrown.

For reference, the `pom.xml` in `spring-ai-openai` includes the `spring-webflux` dependency:
https://github.com/spring-projects/spring-ai/blob/main/models/spring-ai-openai/pom.xml

### Describe how you did it
Added the `spring-webflux` dependency to the `spring-ai-openai` package where `DashScopeApi` is located.

### Describe how to verify it
Users don't need to add the `spring-webflux` dependency manually.

### Special notes for reviews
image

### DIFF
--- a/spring-ai-alibaba-core/pom.xml
+++ b/spring-ai-alibaba-core/pom.xml
@@ -55,6 +55,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-webflux</artifactId>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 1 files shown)


yuluo-yx (approved):
LGTM, thanks @YunKuiLu

PR #1349, opened 2025-06-24
fix(graph): fixed some issues with the current DSL conversion

Wangzy455:
1. Added support for the startNode node in dify and added an interface in the controller
2. Added BranchNode nodes to adapt to dify
3. Fixed rendering issues with some nodes and correctly registered variables in overAllState
4. Added missing dependencies

1，添加了对 dify 中 startNode 节点的支持，在 controller 中添加了一个接口
2，添加了 BranchNode 节点对 dify 进行适配
3，修复了部分节点的渲染问题，可以正确的在 overAllState 里注册变量
4，添加了缺失的依赖


spring-ai-alibaba-graph/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/node/BranchNode.java
```diff
@@ -0,0 +1,71 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.cloud.ai.graph.node;
+
+import com.alibaba.cloud.ai.graph.OverAllState;
+import com.alibaba.cloud.ai.graph.action.NodeAction;
+import java.util.*;
```
chickenlj: Please avoid using asterisk-style import.

spring-ai-alibaba-graph/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/node/BranchNode.java
```diff
@@ -0,0 +1,71 @@
...
+	public BranchNode(String outputKey, String inputKey) {
+		this.inputKey = inputKey;
+		this.outputKey = outputKey;
+	}
+
+	@Override
+	public Map<String, Object> apply(OverAllState state) throws Exception {
+		Map<String, Object> updatedState = new HashMap<>();
+		// use the inputKey to get the value
+		if (StringUtils.hasLength(this.inputKey)) {
```
chickenlj: It seems that `inputKey` and `outputKey` are mandatory for BranchNode to work properly. So I think we should add assertion checks in Constructor or Builder.

spring-ai-alibaba-graph/spring-ai-alibaba-graph-studio/src/main/resources/templates/GraphRunController.java.mustache
```diff
@@ -34,5 +35,10 @@ public class GraphRunController {
         return graph.invoke(inputs).orElse(null);
     }
 
+    @PostMapping(value = "/start")
+    public Map<String, Object> startInvoke(@RequestBody Map<String, Object> inputs) throws Exception {
+        {{startInputsSection}}
```
chickenlj, line 40: What's the difference between `/start` and `/invoke`?  Could you please give an example of the actual generated value of `{{startInputsSection}}`?

Wangzy455: At that time, the idea here was that there is no actual definition node for this startNode in the graph, so it is not possible to directly access the variables in startNode from dify. Therefore, it is equivalent to extracting predefined parameters from the dsl of dify for use.

HY-love-sleep: > What's the difference between `/start` and `/invoke`? Could you please give an example of the actual generated value of `{{startInputsSection}}`?

startNode is currently stateless in the graph, just an identification node, but in dify dsl, the startNode will store some initialization variables, here for compatibility, put the initialization variable in startNode in difyDSL into the invoke parameter
image

Wangzy455: image
When the initial node has only one parameter, it is similar to this.

spring-ai-alibaba-graph/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/node/BranchNode.java
Wangzy455: get

spring-ai-alibaba-graph/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/node/BranchNode.java
Wangzy455: get

spring-ai-alibaba-graph/spring-ai-alibaba-graph-studio/src/main/resources/templates/GraphRunController.java.mustache
HY-love-sleep: > image When the initial node has only one parameter, it is similar to this.

But I think this should be
startInputs.put("useLLM", "yes"));  or startInputs.put("useLLM", "no")); 
This value should be taken directly from the DSL, Because it is a required field in the DSL

HY-love-sleep: > > image When the initial node has only one parameter, it is similar to this.
> 
> But I think this should be startInputs.put("useLLM", "yes")); or startInputs.put("useLLM", "no")); This value should be taken directly from the DSL, Because it is a required field in the DSL

sorry， It seems that this will have to be put in the body of startInvoke for user-defined input, but you can add some annotations to constrain the user from assigning values to the necessary fields

PR #1547, opened 2025-07-10
perf(deepresearch): Optimize reporter node prompts

hbsjz-swl:
Refactored the reporter node system role positioning prompts, It seems that the output report is a bit thicker

### Special notes for reviews
Before：
15740fe800f8569c359b5d782751b4b8
After:
4c44f9b66799457f02c1d5556b1d6094


hbsjz-swl:
The effect of using deepseek-r10528 model is as follows:
image

GTyingzi (approved):
LGTM

PR #1561, opened 2025-07-11
feat(deepresearch): adding user-defined allowlists and blocklists of website

VLSMB:

Sort and filter search engine query results to reduce the impact of incorrect information.

NONE

### Describe how you did it

I defined a `SearchFilterService` class that can sort and filter search results returned by `SearchService` based on user-configured information. To support different configuration sources, I defined the abstract method `abstract Map<String, Double> loadWebsiteWeight()`. This is the only abstract method in the class and must be overridden by derived classes. The method returns a Map where the keys are website domain names and the values are their weights (ranging from -1.0 to 1.0; a negative value indicates distrust, a positive value indicates trustworthiness, and a larger absolute value signifies a higher degree of trust or distrust).

This class has two query methods, both of which return a search list sorted in descending order of website weight. The only difference between them is whether they retain search results with negative weights. Currently, both `BackgroundInvestigationNode` and `ResearcherNode` use the version that does not retain search results with negative weights.

Because the `Deepresearch` framework needs to record the search results from `ChatClient` calling the search engine within `ResearcherNode`, I additionally defined a `SearchFilterTool` passed to `ChatClient`.


spring-ai-alibaba-deepresearch/src/main/java/com/alibaba/cloud/ai/example/deepresearch/service/LocalConfigSearchFilterService.java
```diff
@@ -0,0 +1,68 @@
...
+		super(searchBeanUtil);
+	}
+
+	@Override
+	protected Map<String, Double> loadWebsiteWeight() {
+		ClassPathResource resource = new ClassPathResource("website-weight-config.json");
+		Map<String, Double> map = new HashMap<>();
+		try (InputStream stream = resource.getInputStream()) {
+			List<WebsiteConfig> configs = objectMapper.readValue(stream,
+					objectMapper.getTypeFactory().constructCollectionType(List.class, WebsiteConfig.class));
```
mengnankkkk, line 54: 对配置文件的读取加一层校验，避免出现非法值。

spring-ai-alibaba-deepresearch/src/main/java/com/alibaba/cloud/ai/example/deepresearch/service/LocalConfigSearchFilterService.java
```diff
@@ -0,0 +1,68 @@
...
+			configs.forEach(config -> map.put(config.host(), config.weight()));
+		}
+		catch (Exception e) {
+			log.error("Failed to read website weight configuration file: {}", e.getMessage());
```
mengnankkkk: 直接抛出异常，可能导致服务不可用。可以返回一个空的map，然后进行告警

VLSMB: Thank you for your suggestion. I have fixed the relevant code

PR #1612, opened 2025-07-14
feat(nl2sql): support running python code within docker containers, and the containers can be reused

VLSMB:

The nl2sql module sometimes needs to execute Python code for analysis.

NONE

### Describe how you did it

Similar to a thread pool, I've defined an `AbstractContainerPoolExecutor` class to manage container creation and usage. Containers are categorized into core containers and temporary containers. Their quantities, along with the destruction time of temporary containers, are user-configurable.

When a new code task arrives, the system prioritizes using existing available containers. If none are available, it checks whether the container count has exceeded the combined limit of core and temporary containers. If within limits, a new container is created; otherwise, the task enters a queue to await available capacity.

This design enables parallel execution of multiple code tasks while maximizing container reuse, thus eliminating performance overhead from frequent container creation/destruction. Each code task record contains the code itself, standard input, and third-party dependencies – all parameters that will be provided by the AI model during tool invocation.

### Describe how to verify it

In the unit test class, validate its availability by executing task threads that exceed the total container capacity.

### Special notes for reviews

Currently, only the Docker implementation class has been completed. If this PR can be merged, I will proceed to implement the classes for Kata and Containerd.


zhangshenghang:
thanks @VLSMB , It looks great,  can a test class be added to verify it? For example, we use a sentence and data as input parameters to automatically verify the feasibility of this tool.

VLSMB:
> thanks @VLSMB , It looks great, can a test class be added to verify it? For example, we use a sentence and data as input parameters to automatically verify the feasibility of this tool.

Hello, There is already a basic unit test class for the core code in `spring-ai-alibaba-nl2sql-management/src/test/java/com/alibaba/cloud/ai/controller/DockerContainerPoolExecutorTest.java`. I will later supplement it with a more comprehensive integration test.

spring-ai-alibaba-nl2sql/spring-ai-alibaba-nl2sql-chat/src/main/java/com/alibaba/cloud/ai/node/PythonExecuteNode.java
zhangshenghang, line 1: Currently, directly referencing this node will prevent the process from running smoothly as there are still some tasks that haven't been addressed. Should we first modify it to PythonExecuteNodeV2 and then add it to the process when it is fully available?

VLSMB: Hello, I've already finished the modifications.

zhangshenghang (approved):

PR #1954, opened 2025-08-01
fix(tool-calls): broken trace of tool call

Cirilla-zmh:

Fix the broken trace of tool call.


flystar32 (approved):
LGTM

PR #1962, opened 2025-08-02
fix(nl2sql): fix incorrect foreign key creation order in SQL table ge…

ShlyxCoder:

Fixes the issue in the original nl2sql module where the order of external references during table creation was incorrect.
修复图

### Describe how you did it
Adjusted the execution order of the sql logic.

### Describe how to verify it
Compare the SQL generated by the previous version and the current version to confirm that the execution order has been corrected.


VLSMB (approved):
LGTM

PR #1987, opened 2025-08-04
refactor(nl2sql): replace @Autowired with constructor injection for b…

zxuexingzhijie:


spring-ai-alibaba-nl2sql/spring-ai-alibaba-nl2sql-management/src/main/java/com/alibaba/cloud/ai/controller/AgentController.java
```diff
@@ -32,9 +31,12 @@
 @CrossOrigin(origins = "*")
 public class AgentController {
 
-	@Autowired
 	private AgentService agentService;
```
zhangshenghang: Uniformly change it to `final`, and the same applies later.

zhangshenghang (approved):

PR #2026, opened 2025-08-06
[WIP] feat(observation): Support OpenTelemetry semantic convention of input & output

Cirilla-zmh:


clongbupt:
To keep a record of input/output messages, consider extending DefaultChatModelObservationConvention and overriding getHighCardinalityKeyValues, then add new key-value pairs, may be a simpler way.

PR #2233, opened 2025-08-23
feat(nl2sql): enhance python code generate and display

VLSMB:

1. Added a locally executable implementation class in the Python code and designed a more effective Prompt.  
2. Implemented Python syntax highlighting in the frontend.  
3. Fixed several frontend bugs: the send button was unresponsive; when re-executing a Node (SQL or Python generation), the output was appended to the previous result's HTML block, but it should instead display in a new HTML code block.


zhangshenghang:
Thank you. I will merge it in the near future.

zhangshenghang (approved):

Issue #1262, opened 2025-06-17
[BUG] <title>isStreamingToolFunctionCallFinish 方法 output为null 报空指针异常


swit1983:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

大模型对话流式输出

### Environment

```markdown
Spring AI Alibaba version(s):1.0.6M
```

### Debug logs

`java.lang.NullPointerException: Cannot invoke "com.alibaba.cloud.ai.dashscope.api.DashScopeApi$ChatCompletionOutput.choices()" because the return value of "com.alibaba.cloud.ai.dashscope.api.DashScopeApi$ChatCompletionChunk.output()" is null
	at com.alibaba.cloud.ai.dashscope.api.DashScopeAiStreamFunctionCallingHelper.isStreamingToolFunctionCall(DashScopeAiStreamFunctionCallingHelper.java:170) ~[spring-ai-alibaba-core-1.0.0-M6.1.jar:1.0.0-M6.1]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Assembly trace from producer [reactor.core.publisher.FluxMapFuseable] :
	reactor.core.publisher.Flux.map(Flux.java:6631)
	com.alibaba.cloud.ai.dashscope.api.DashScopeApi.chatCompletionStream(DashScopeApi.java:1366)
Error has been observed at the following site(s):
	*________________Flux.map ⇢ at com.alibaba.cloud.ai.dashscope.api.DashScopeApi.chatCompletionStream(DashScopeApi.java:1366)
	|_       Flux.windowUntil ⇢ at com.alibaba.cloud.ai.dashscope.api.DashScopeApi.chatCompletionStream(DashScopeApi.java:1372)
	|_ Flux.concatMapIterable ⇢ at com.alibaba.cloud.ai.dashscope.api.DashScopeApi.chatCompletionStream(DashScopeApi.java:1379)
	|_           Flux.flatMap ⇢ at com.alibaba.cloud.ai.dashscope.api.DashScopeApi.chatCompletionStream(DashScopeApi.java:1384)
	|_               Flux.map ⇢ at com.alibaba.cloud.ai.dashscope.chat.DashScopeChatModel.lambda$stream$12(DashScopeChatModel.java:250)
[...]

PR #2253, opened 2025-08-26
fix(core): check chat completion chunk

cxhello:

Fixes https://github.com/alibaba/spring-ai-alibaba/issues/1262


VLSMB (approved):
LGTM

Issue #2305, opened 2025-09-02
[BUG] JedisRedisChatMemoryRepository message media 序列化异常


liusishan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### **异常描述：**

使用 JedisRedisChatMemoryRepository 实现对话记忆的功能，在使用视觉多模态模型(Vision)时，
Media 对象序列化失败：当 data 不是 byte[] 时，会抛出 JsonMappingException

### 原因：
- Media#getDataAsByteArray() 仅在 data 为 byte[] 时才能返回成功，否则会抛出 IllegalStateException。
- Jackson 在序列化 Media 时会自动调用 getDataAsByteArray()（因为属性名匹配），导致抛出异常。

**异常如下:**
```
ERROR o.springframework.ai.chat.model.MessageAggregator - Aggregation Error
java.lang.RuntimeException: Error serializing message
	at com.alibaba.cloud.ai.memory.redis.BaseRedisChatMemoryRepository.serializeMessage(BaseRedisChatMemoryRepository.java:63)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
	at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
	at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)
	at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:622)
	at java.base/java.util.stream.ReferencePipeline.toList(ReferencePipeline.java:627)
	at com.alibaba.cloud.ai.memory.redis.JedisRedisChatMemoryRepository.saveAll(JedisRedisChatMemoryRepository.java:198)
	at org.springframework.ai.chat.memory.MessageWindowChatMemory.add(MessageWindowChatMemory.java:65)
	at org.springframework.ai.chat.memory.ChatMemory.add(ChatMemory.java:46)
[...]

PR #2324, opened 2025-09-04
fix(deepresearch): fix deserialize in deepresearch and memcached

benym:
The same reason as #2305

### Special notes for reviews
@VLSMB


VLSMB:
Hello, the test failed. You likely need to configure the `ObjectMapper` in the `MemcachedChatMemoryRepository`. And rename the unit test class to end with 'Test'.

image

benym:
> Hello, the test failed. You likely need to configure the `ObjectMapper` in the `MemcachedChatMemoryRepository`. And rename the unit test class to end with 'Test'.
> 
> image

done，another change in DeepResearchStateSerializer

PR #2348, opened 2025-09-08
chore(graph): remove unless jboss logging

benym:

A conflict between jboss and logback will cause SpringBootTest to report an error

jboss

### DIFF
--- a/spring-ai-alibaba-graph-core/pom.xml
+++ b/spring-ai-alibaba-graph-core/pom.xml
@@ -47,6 +47,24 @@
             <groupId>io.github.a2asdk</groupId>
             <artifactId>a2a-java-reference-server</artifactId>
             <version>${a2a-sdk.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jboss.threads</groupId>
+                    <artifactId>jboss-threads</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.jboss.logmanager</groupId>
+                    <artifactId>jboss-logmanager</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.jboss.logging</groupId>
+                    <artifactId>jboss-logging</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.jboss.slf4j</groupId>
+                    <artifactId>slf4j-jboss-logmanager</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.github.a2asdk</groupId>


VLSMB:
Hello, could you tell me which unit test class this error is from?

benym:
> Hello, could you tell me which unit test class this error is from?

For the time being, it has not affected the existing unit test code in the repository. However, any project that relies on `spring-ai-alibaba-graph-core` will report this issue when writing unit tests containing SpringBootTest. The smallest reproduction step only requires adding such a line to the existing single test

image

The default jboss dependency introduction will force users of spring-ai-alibaba-graph-core to exclude conflicting packages themselves

VLSMB:
I'm wondering if it's sufficient to only exclude the `slf4j-jboss-logmanager` dependency. I'm uncertain whether the project or any of its imported dependencies are using other JBoss logging components, so I'm concerned that excluding everything might carry some risks.

benym:
> I'm wondering if it's sufficient to only exclude the `slf4j-jboss-logmanager` dependency. I'm uncertain whether the project or any of its imported dependencies are using other JBoss logging components, so I'm concerned that excluding everything might carry some risks.

Simply excluding `slf4j-jboss-logmanager` does indeed meet current compatibility requirements. I understand your concerns, but after further investigation, I discovered a more significant issue: `a2a-java-reference-server` was renamed from `a2a-java-sdk-server-quarkus` in the `a2aproject` open source repository. In the upcoming `v0.3.0.Alpha1` release, `a2a-java-reference-server` will be renamed to `a2a-java-sdk-reference-jsonrpc`. https://github.com/a2aproject/a2a-java/releases. The SDK name appears to be unstable. I downloaded the source code and reviewed the current version's core dependencies, and found that the `a2a-java-reference-server` implementation depends on `Quarkus`. `JBoss` is introduced through this dependency. **I consulted the Quarkus official manual, which recommends excluding other logging framework implementations and retaining only the JBoss framework, as JBoss is the entire Quarkus logging platform.** 
https://docs.redhat.com/zh-cn/documentation/red_hat_build_of_quarkus/3.8/html/logging_configuration/logging-apis
Since the document clearly states that `slf4j-jboss-logmanager` should be introduced when using `SLF4J`, it is not practical to directly exclude `slf4j-jboss-logmanager`, this may result in errors such as `ClassNotFoundException`, `NoSuchMethodError`, `NoClassDefFoundError` at runtime

What I want to say more is that we all know that `Quarkus` and `SpringBoot` are **two different microservice frameworks**, but when we want to implement the A2A function in `SpringBoot`, we need to rely on `Quarkus`, which is very strange. And if jboss is completely removed, I am not sure whether there will be some problems with Quarkus that is dependent on a2a sdk, but as long as `spring-ai-alibaba-graph-core` includes `a2a-java-reference-server`, jboss logging framework will conflict with springboot's native logback. I think the a2a sdk should be moved to the module to which a2a belongs to avoid frequent conflicts caused by users not using the a2a function. In addition, the implementation of a2a needs to be reconsidered, including the possible merging of pr in the future #2332, besides, I haven't seen any test cases related to a2a yet. The original `RemoteAgentTest` test section was removed in #2296, which makes me very worried that the a2a functionality might cause conflicts when integrated into large existing projects. However, the conflicts themselves stem from Quarkus and SpringBoot

VLSMB:
Hello, thank you for your suggestion. I have already relayed this issue to the developers responsible for this part. It might require an alternative approach to resolve it.

PR #2384, opened 2025-09-11
refactor(studio): improve date parsing error handling

6mmLIU:
Replaces calls to printStackTrace with an SLF4J logger in DateUtils.parseDateString so that failures are logged appropriately without throwing uncaught exceptions. On parse failure, the method now logs a warning and returns null. Adds unit test coverage for both successful parsing and failure scenarios to prevent regressions and ensure robustness.

NONE

### Describe how you did it

Introduced a private static SLF4J Logger in DateUtils.

Wrapped the sdf.parse(time) call in a try/catch block; on ParseException, call logger.warn("Failed to parse date string: {}", time, e) and return null.

Wrote JUnit tests verifying that:

A valid date string (e.g. "Wed Sep 10 14:23:45 GMT 2025") returns a non-null Date.

An invalid string (e.g. "not-a-date") yields null and emits a warning.

### Describe how to verify it
Run the core module’s tests:

mvn -q -pl spring-ai-alibaba-studio/spring-ai-alibaba-studio-server/spring-ai-alibaba-studio-server-core -am test

Verify all tests pass. Additionally, trigger a failure case in a debugger or by running a small main method to confirm that a warning is logged (rather than a stack trace printed) when parsing an invalid string.


VLSMB:
Please format your code and have a LICENSE before each new file is added

tools/bin/gitleaks
VLSMB, line 1: This file is probably not needed

PR #2483, opened 2025-09-22
fix(core): change config AI_DASHSCOPE_API_KEY to DASHSCOPE_API_KEY

co63oc:
修改配置AI_DASHSCOPE_API_KEY 为 DASHSCOPE_API_KEY，网页上为 DASHSCOPE_API_KEY
如果CI设置了环境变量，可能CI流水线也需要修改环境变量
image


chickenlj:
This change might bring incompatibilities for users. So I suggest doing it in the next major release.

co63oc:
> This change might bring incompatibilities for users. So I suggest doing it in the next major release.

我也不确定那看项目安排

Issue #2467, opened 2025-09-20
[BUG] embedding失败


czlh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

@Configuration
public class FastJsonMvcConfig {

    @Bean
    public HttpMessageConverters fastJsonHttpMessageConverters() {
        FastJsonHttpMessageConverter converter = new FastJsonHttpMessageConverter();
        FastJsonConfig config = new FastJsonConfig();
        config.setDateFormat("yyyy-MM-dd HH:mm:ss");
        config.setWriterFeatures(
                JSONWriter.Feature.WriteMapNullValue,
                JSONWriter.Feature.WriteLongAsString,
                JSONWriter.Feature.WriteNullStringAsEmpty,
                JSONWriter.Feature.PrettyFormat
        );
        converter.setFastJsonConfig(config);
        converter.setDefaultCharset(StandardCharsets.UTF_8);
        return new HttpMessageConverters(converter);
    }
}

加上上面的配置后，embedding失败，主要是JSONWriter.Feature.WriteNullStringAsEmpty配置项

HTTP 400 - {"request_id":"6482c6aa-29a2-4489-9b62-16ae15f9d317","code":"InvalidParameter","message":"<400> InternalError.Algo.InvalidParameter: Value error, text_type is invalid, its value should be in ['document', 'query'].: parameters.text_type"}

### Environment

```markdown
Spring AI Alibaba version(s):
```

PR #2517, opened 2025-09-25
fix: Handle null textType in DashScope embedding API for FastJson compatibility

Libres-coder:
This PR resolves a compatibility issue between the DashScope embedding API and FastJson serialization configuration when the textTypefield is null.When users enable FastJson's WriteNullStringAsEmptyfeature, a nulltextTypefield is serialized into an empty string (""), causing the DashScope API call to fail with the error: "Parameter error, text_type should be query or document".

This PR addresses a specific issue: the compatibility problem where FastJson's WriteNullStringAsEmptyconfiguration causes embedding API calls to fail. https://github.com/alibaba/spring-ai-alibaba/issues/2467

### Describe how you did it
By adding defensive handling logic in the DashScopeApi.EmbeddingRequestInputParameters.Builder.build() method:
"""
public EmbeddingRequestInputParameters build() {
// Handle null textType for FastJson compatibility.
// When FastJson WriteNullStringAsEmpty is enabled, null textType becomes empty string, causing API errors.
String finalTextType = textType == null ? DEFAULT_EMBEDDING_TEXT_TYPE : textType;
return new EmbeddingRequestInputParameters(finalTextType, dimension);
}
"""
When textType is null, it is automatically set to the default value ("document"), ensuring API call stability.


Libres-coder:
@zxuexingzhijie  Could you please review my PR?

Libres-coder:
@zxuexingzhijie Could you please review my PR?

Issue #2402, opened 2025-09-15
[BUG]


hunk666:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

调用知识库删除方法报错
控制台报错如下：
    
<img width="2409" height="562" alt="Image" src="https://github.com/user-attachments/assets/e7d27f12-f6f1-416a-86db-2f2f5ce80ce9" />
返回：{"code":"SystemError","status_code":500,"message":"An system error has occured, please try again later.","request_id":"3a7897af-ee9b-4f89-b18c-c2bc7ad480f9"}

提工单了，工程师回复说是参数传错了：

<img width="1504" height="1094" alt="Image" src="https://github.com/user-attachments/assets/950004f7-cf8d-4d2b-8ed0-e2790df267b1" />

### Steps To Reproduce

`<dependency>
                <groupId>com.alibaba.cloud.ai</groupId>
                <artifactId>spring-ai-alibaba-bom</artifactId>
                <version>1.0.0.3</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
`

`package com.winco.ai.service.rag;

import com.alibaba.cloud.ai.advisor.DocumentRetrievalAdvisor;
import com.alibaba.cloud.ai.dashscope.api.DashScopeApi;
import com.alibaba.cloud.ai.dashscope.chat.DashScopeChatModel;
import com.alibaba.cloud.ai.dashscope.embedding.DashScopeEmbeddingOptions;
import com.alibaba.cloud.ai.dashscope.rag.*;
import lombok.extern.slf4j.Slf4j;
import org.springframework.ai.chat.client.ChatClient;
import org.springframework.ai.chat.client.advisor.SimpleLoggerAdvisor;
import org.springframework.ai.chat.prompt.SystemPromptTemplate;
import org.springframework.ai.document.Document;
import org.springframework.ai.document.DocumentReader;
import org.springframework.ai.rag.retrieval.search.DocumentRetriever;
import org.springframework.ai.vectorstore.VectorStore;
import org.springframework.beans.factory.annotation.Value;
import org.springframework.core.io.Resource;
import org.springframework.stereotype.Service;
import reactor.core.publisher.Flux;

import java.io.File;
import java.io.FileOutputStream;
import java.io.IOException;
import java.io.InputStream;
import java.util.List;
import java.util.stream.Collectors;

@Slf4j
@Service
public class DashscopeRagServiceImpl {

[...]

PR #2531, opened 2025-09-28
fix(core): fix the dashscope cloud store delete

Fancy-hjyp:
The type of param `component` should be object not list in `com.alibaba.cloud.ai.dashscope.api.DashScopeApi#deletePipelineDocument`, causing  the document to fail to be deleted.

Fixes https://github.com/alibaba/spring-ai-alibaba/issues/2402

image

### Describe how you did it
Change the parameter to the correct type.


Fancy-hjyp:
@yuluo-yx Please take  a look, thanks.

PR #2547, opened 2025-09-30
fix(mcp): fix mcp router switch to default close 

Aias00:
fix mcp router switch to default close


yuluo-yx (approved):
LGTM

PR #2566, opened 2025-10-10
fix(graph): replace blocking CompletableFuture.join() with async callbacks in toFlux method

### LINKED ISSUE #2559: [BUG] There is a performance issue with the private <E> Flux<GraphResponse<E>> toFlux(AsyncGenerator generator) function in A2aNodeWithConfig

### Current Behavior

The toFlux function on line 100 of the file uses CompletableFuture.join() cyclically, which can cause a large number of threads to be blocked and occupy resources. In absolute scenarios, this can lead to abnormal server resource consumption. In addition, the official website has given a clear exception for the exception thrown by the join() method: https://docs.oracle.com/javase/8/docs/api/

```
private <E> Flux<GraphResponse<E>> toFlux(AsyncGenerator generator) {
		return Flux.create(sink -> {
			Schedulers.boundedElastic().schedule(() -> {
				try {
					AsyncGenerator.Data<E> data;
					while (!(data = generator.next()).isDone()) {
						if (data.isError()) {
							// Handle error case
							try {
								data.getData().join(); // This will throw the exception
							}
							catch (Exception e) {
								sink.error(e);
								return;
							}
						}
						else {
							// Emit the element
							try {
								E element = data.getData().join();
								sink.next(GraphResponse.of(element));
							}
							catch (Exception e) {
								sink.error(e);
								return;
							}
						}
					}

					if (data.resultValue().isPresent()) {
						sink.next(GraphResponse.done(data.resultValue().get()));
					}
					sink.complete();
				}
				catch (Exception e) {
					sink.error(e);
				}
			});
		});
	}
```

SaladDay:

This PR fixes a critical performance issue in the `toFlux()` method of `A2aNodeActionWithConfig` class where `CompletableFuture.join()` was being used cyclically, causing thread blocking and resource consumption issues.

Fixes #2559

### Describe how you did it

- Converted synchronous `join()` calls to asynchronous `whenComplete()` callbacks
- Implemented proper `Disposable` handling for stream cancellation
- Split complex logic into a dedicated `drainGenerator()` method using recursive approach
- Added `unwrapCompletionException()` method for better exception handling

### Describe how to verify it
```
./mvnw -pl spring-ai-alibaba-graph-core test -Dtest=com.alibaba.cloud.ai.graph.agent.a2a.A2aNodeActionWithConfigTests
```
- `toFluxEmitsAsyncData()`: Verifies normal async data emission
- `toFluxPropagatesErrors()`: Verifies proper error propagation
-  All existing functionality preserved
-  Performance improvement: No more thread blocking on `join()` calls

### Special notes for reviews

This change maintains backward compatibility while significantly improving performance by eliminating blocking operations. The recursive approach ensures proper async processing without resource leaks.

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/agent/a2a/A2aNodeActionWithConfig.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/agent/a2a/A2aNodeActionWithConfig.java
@@ -50,7 +50,9 @@ import org.apache.http.entity.StringEntity;
 import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
 import org.apache.http.util.EntityUtils;
+import reactor.core.Disposable;
 import reactor.core.publisher.Flux;
+import reactor.core.publisher.FluxSink;
 import reactor.core.scheduler.Schedulers;
 
 import static java.lang.String.format;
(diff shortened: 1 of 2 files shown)


SaladDay:
ptal, thx💕💕

PR #2636, opened 2025-10-22
fix(a2a): ReactAgent instruction changes not taking effect immediately

Sunrisea:
Fix ReactAgent instruction changes not taking effect immediately


Sunrisea:
@chickenlj  Please check it

PR #2648, opened 2025-10-24
feat(agent): add Structured output

logicwu0:


chickenlj:
The overall logic seems sound, with a few key points:
1. Implementation should be made against the agent-framework module and should be based on `AgentLlmNode` and `ReactAgent`.
2. The `format_output` tool has not been set
3. The processing logic for response values may also need adjustment in conjunction with `format_output` tool
4. Finally, consider using an independent `state key` to store the parsed response value

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/structured/StructuredOutputIntegration.java
```diff
@@ -0,0 +1,187 @@
...
+        try {
+            schema = parseJsonSchema(outputSchema);
+        } catch (Exception e) {
+            return null;
+        }
+
+        if (mode == null) {
+            String provider = inferProvider(modelName);
+            mode = StructuredOutputSupport.detectMode(provider, null);
+        }
```
chickenlj, line 114: I think we forget to set the `responseFormat` of `model options`  for NATIVE mode.

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/structured/StructuredOutputIntegration.java
```diff
@@ -0,0 +1,187 @@
...
+        ToolCallback formatOutputTool = prepareStructuredOutput(outputSchema, null);
+
+        if (formatOutputTool != null) {
+            List<ToolCallback> allTools = new ArrayList<>(existingToolCallbacks);
+            allTools.add(formatOutputTool);
+
+            return ToolCallingChatOptions.builder()
+                .toolCallbacks(allTools)
+                .internalToolExecutionEnabled(false)
+                .build();
```
chickenlj: Make sure copy all the fields of the previous ToolCallingChatOptions instance.

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/structured/StructuredOutputIntegration.java
logicwu0: Thanks! You’re right — I forgot to set the responseFormat before. I’ve fixed it in the new commit.

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/structured/StructuredOutputIntegration.java
logicwu0: Thanks! I’ve updated the implementation to copy all fields from the previous ToolCallingChatOptions instance.

PR #2678, opened 2025-11-01
fix(a2a): fix the issue of recursive bean creation

Wangzy455:
Fix the issue of recursive bean creation


chickenlj (approved):
LGTM.

PR #2701, opened 2025-11-06
fix(agent): resolve ReactAgent ThreadLocal safety in concurrent scenarios

JGoP-L:

This PR enhances thread safety for the `ReactAgent` class in the agent framework module by fixing a critical concurrency issue with the iteration counter. 

**Problem:**
When `ReactAgent` instances are used as singleton Spring beans (which is the common pattern in Spring applications, as seen in `AgentStaticLoader`), the instance variable `private int iterations` was shared across all concurrent requests. This caused:
1. Race conditions where multiple threads increment the same counter simultaneously
2. Iteration count accumulation across sequential calls (counter never resets)
3. Unpredictable agent behavior including premature termination or exceeding max iterations
4. Memory visibility issues in multi-threaded environments

**Solution:**
Refactored the iteration counter to use `ThreadLocal<Integer>` to ensure thread isolation, with proper lifecycle management (reset and cleanup) at all entry points (`doInvoke` and `doStream` methods).

NONE

### Describe how you did it

**Core Changes:**

1. **Replaced instance variable with ThreadLocal** (`ReactAgent.java`):
   - Changed `private int iterations = 0;` to `private ThreadLocal<Integer> iterations = ThreadLocal.withInitial(() -> 0);`
   - Updated `makeModelToTools()` to use `iterations.get()` and `iterations.set()`

2. **Overrode entry point methods** to manage ThreadLocal lifecycle:
   - `doInvoke(Map<String, Object>, RunnableConfig)`: Reset counter to 0 at start, cleanup in finally block
   - `doStream(Map<String, Object>)`: Reset counter to 0 at start, cleanup with `doFinally()`
   - `doStream(Map<String, Object>, RunnableConfig)`: Same pattern as above

3. **Fixed iteration comparison logic**:
   - Changed `if (iterations++ > maxIterations)` to `if (currentIterations >= maxIterations)` for correct boundary check

**Why this approach:**
- ThreadLocal ensures each thread has its own isolated counter
- Overriding all entry points ensures proper reset/cleanup regardless of invocation method (invoke vs stream)
- Finally blocks prevent ThreadLocal memory leaks even when exceptions occur
- Maintains backward compatibility - no API changes

### Describe how to verify it

**Test Coverage:**

1. **Unit Tests** (`ReactAgentThreadLocalTest.java`):
   - Verifies ThreadLocal field exists via reflection
   - Confirms all entry methods (`doInvoke`, `doStream` variants) are properly overridden
   - Validates that ReactAgent class declares these methods (not inherited)

2. **Concurrency Integration Tests** (`ReactAgentConcurrencyTest.java`):
   - **Concurrent call() method test**: 10 threads × 5 calls each = 50 concurrent invocations
   - **Concurrent invoke() method test**: 10 threads × 5 direct invocations each
   - **Concurrent stream() method test**: 10 threads × 5 streaming calls each
   - **Thread pool reuse test**: 50 tasks in 3-thread pool to verify ThreadLocal cleanup
   - **Mixed method test**: Concurrent calls to `call()`, `invoke()`, and `stream()` simultaneously

All tests verify:
- No `ConcurrentModificationException` or race conditions
- Each invocation gets correct iteration count (starts from 0)
- Results are consistent across all threads
- No counter contamination between threads or sequential calls

**Manual Verification:**
```bash
# Run all tests in the agent framework module
mvn test -pl spring-ai-alibaba-agent-framework

# Run only the new test classes
mvn test -Dtest=ReactAgentThreadLocalTest
mvn test -Dtest=ReactAgentConcurrencyTest -DAI_DASHSCOPE_API_KEY=your_key
```

### Special notes for reviews

**Important Context:**

1. **Why ThreadLocal instead of AtomicInteger?**
   - AtomicInteger only solves the race condition for incrementing
   - It does NOT solve the accumulation problem (counter never resets between calls)
   - ThreadLocal provides complete isolation per thread/request

2. **Why override all entry methods?**
   - `BaseAgent.invoke()` → `doInvoke()` → graph execution
   - `BaseAgent.stream()` → `doStream()` → graph streaming
   - Without overriding, some code paths would miss the reset/cleanup logic

3. **Spring Bean Lifecycle:**
   - In Spring, beans are singletons by default
   - `AgentStaticLoader` creates ReactAgent instances in constructor and reuses them
   - This makes the thread-safety fix essential for production use

4. **Test Requirements:**
   - Concurrency tests require API key: `AI_DASHSCOPE_API_KEY` environment variable
   - Tests use `@EnabledIfEnvironmentVariable` to skip when key is not available
   - All tests follow project standards (public methods, static imports, JavaDoc)

5. **Backward Compatibility:**
   - No public API changes
   - Existing code continues to work without modifications
   - Only internal implementation changed for thread safety

**Review Checklist:**
- ✅ Thread safety verified with comprehensive concurrency tests
- ✅ Memory leaks prevented with proper ThreadLocal cleanup
- ✅ All entry points covered (invoke + stream variants)
- ✅ Tests conform to project standards
- ✅ No breaking changes to public API
- ✅ Documentation updated (JavaDoc comments)

### DIFF
--- a/spring-ai-alibaba-graph-core/pom.xml
+++ b/spring-ai-alibaba-graph-core/pom.xml
@@ -41,6 +41,7 @@
         <testcontainers.version>1.19.3</testcontainers.version>
         <httpclient.version>4.5.14</httpclient.version>
         <jackson.version>2.18.4</jackson.version>
+        <jetbrains-annotations.version>13.0</jetbrains-annotations.version>
     </properties>
 
     <dependencies>
@@ -85,7 +86,7 @@
         <dependency>
             <groupId>org.jetbrains</groupId>
             <artifactId>annotations</artifactId>
-            <version>13.0</version>
+            <version>${jetbrains-annotations.version}</version>
         </dependency>
 
         <dependency>


JGoP-L:
@zxuexingzhijie @disaster1-tesk @yuluo-yx @Aias00 ptal, thx

chickenlj:
Thanks for the contribution. We have removed the iterations field in ReactAgent, and it's been replaced with ModelCallLimitHook.

PR #3133, opened 2025-11-24
Refactor[graph]: optimize keyStrategy management and usage in CompiledGraph and OverallState

chickenlj:
This fixes:
1. ClassCastException when using Agent.asNode in Graph
2. Make sure the parent Graph always has a full State view of all nodes (whether it's normal node, sub state graph, sub compiled graph, ReactAgent).


yuluo-yx:
local test pass, merged.

PR #3135, opened 2025-11-24
fix(graph): support node return "Flux<? extends Nodeoutput>"

chickenlj:


disaster1-tesk:
ci check需要处理一下 @chickenlj

PR #3207, opened 2025-11-27
test(graph): delete `addParallelNodeExecutor("A", ForkJoinPool.commonPool()` to make the test case more suitable for the newly added default executor feature

Zwen092:
…Pool()` to make the test case more suitable for the newly added default executor feature

This PR deletes the use of  `ForkJoinPool.commonPool()` in `testParallelNodeLifecycleListenerNodeId` test case, the default executor implementation is a better choice. Still use commonPool is unnecessary and may cause confusion for new developers who want to learn from unit test.

NONE

### Describe how you did it
just delete the code

### Describe how to verify it
Re-run the unit test and it passed


disaster1-tesk:
LGTM

PR #3138, opened 2025-11-24
fix(graph): preserve toolCalls when merging streaming ChatResponse objects

PeatBoy:
### Describe what this PR does / why we need it

This PR fixes a critical bug in the Flux streaming processing logic where tool calls fail to execute properly, causing graph execution to stop prematurely.

**Root Cause Analysis:**
1. When the AI-generated Flux ChatResponse sequence is `text -> tool -> text`, the final cached result is of text type rather than tool type
2. Subsequently, the `makeModelToTools` method in ReactAgent cannot route to the tool node, naturally failing to trigger the tool callback logic

**Solution:**
Rewrote the logic to properly merge `toolCalls` while merging text content, ensuring tool calls are preserved throughout the streaming process and can be correctly routed by ReactAgent.

### Does this pull request fix one issue?

Fixes the issue where tool calls are lost during Flux streaming, preventing proper tool routing and execution in ReactAgent workflows.

### Describe how you did it

1. Added a new `mergeToolCalls()` method that intelligently merges tool calls from both last and current messages
2. Uses `LinkedHashMap` to preserve tool call order while avoiding duplicates based on tool call ID
3. Modified the message building logic in `getEmbedFlux()` to call `mergeToolCalls()` when constructing the new `AssistantMessage`
4. This ensures that when processing streaming responses with sequence `text -> tool -> text`, all tool calls are properly preserved and available for routing by ReactAgent's `makeModelToTools` method

**Key Changes:**
- Added `mergeToolCalls(List<ToolCall> lastToolCalls, List<ToolCall> currentToolCalls)` method
- Updated `AssistantMessage.builder()` to use merged tool calls: `.toolCalls(mergeToolCalls(lastResponse.getResult().getOutput().getToolCalls(), currentMessage.getToolCalls()))`

### Describe how to verify it

1. Create a test scenario where AI generates a streaming response with sequence: `text -> tool -> text`
2. Verify that tool calls are properly extracted and cached in the final message
3. Verify that ReactAgent's `makeModelToTools` method can correctly route to the tool node
4. Verify that tool callback logic is successfully triggered and executed
5. Verify that graph execution continues properly after tool execution
6. Local testing has confirmed the fix resolves the issue

### Special notes for reviews

- The fix maintains backward compatibility with existing streaming logic
- Tool calls are now preserved across the entire streaming sequence, ensuring proper routing in ReactAgent
- The `LinkedHashMap` ensures both order preservation and deduplication of tool calls

PR #3283, opened 2025-11-28
fix(agent,graph): handle streaming multi-generation assistant me…

zengchao304:
### Summary

  This PR fixes how assistant messages are selected in both non‑streaming and streaming paths when ChatResponse contains multiple Generation objects. It ensures that:

  - Tool calls are not ignored when they appear in a non‑first generation.
  - Usage‑only streaming chunks (with null results/outputs) are skipped safely.
  - Agent‑level code (AgentLlmNode) and graph‑core streaming (NodeExecutor, GraphFluxGenerator) use a consistent assistant‑message selection strategy.

  ### Background / Problem

  Some providers return ChatResponse objects with multiple Generation entries for a single model invocation. At the same time:

  - ChatResponse#getResult() returns only the first generation.
  - Tool calls (AssistantMessage.hasToolCalls() == true) may be present in a later generation instead of the first one.
  - For streaming with usage reporting (e.g. streamUsage), providers may emit usage‑only chunks where:
      - ChatResponse.getResult() is null, or
      - Generation.getOutput() is null.

  Before this change:

  - AgentLlmNode in non‑streaming mode used:

    response.getResult().getOutput()

    which assumes the first generation is always “the real” assistant message.
  - Graph‑core streaming (NodeExecutor.getEmbedFlux and GraphFluxGenerator) also assumed:
      - response.getResult() is non‑null;
      - response.getResult().getOutput() is the right message to stream and aggregate.

  This leads to two main issues:

  1. Tool calls can be ignored.
     When the first generation contains a plain assistant message and a later generation contains the tool call, the system will:
      - Log or return the plain text;
      - Never expose the tool call to the agent/tooling layer.
  2. Streaming can see NullPointerException / inconsistent behavior.
     Usage‑only chunks (with null result/output) are not filtered out early, and downstream code blindly dereferences getResult().getOutput().

  ### Screenshots / Debugging Evidence

  The following debugger screenshots illustrate the problem and how it manifests at runtime:

  1. Non‑streaming path in AgentLlmNode before the fix
     At the breakpoint in AgentLlmNode.apply(...), the code was:

     if (response != null && response.getResult() != null) {
         responseMessage = response.getResult().getOutput();
     }

     The debugger shows that:
      - response.getResults() has two Generation entries.
      - The first generation’s AssistantMessage has no toolCalls.
      - The second generation’s AssistantMessage does contain tool calls.

     However, the code only uses getResult() (the first generation), so the tool call from the second generation is completely ignored.

image

image

  2. Streaming behavior
     A similar situation can arise in streaming: individual ChatResponse chunks may contain multiple generations, and usage‑only chunks may have null outputs. The previous streaming code accessed
     response.getResult().getOutput() directly, which is fragile when:
      - The tool call is in a non‑first generation, or
      - The chunk only carries usage metadata.

  ### What this PR changes

  #### 1. AgentLlmNode (agent framework)

  File:
  spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java

  - Add a helper method:

    private AssistantMessage extractAssistantMessage(ChatResponse response)

    Behavior:
      - Scans all Generation objects via response.getResults().
      - Prefers the first AssistantMessage that has tool calls.
      - If none have tool calls, falls back to the last non‑null AssistantMessage.
      - If getResults() fails or is empty, falls back to response.getResult().getOutput() if it is an AssistantMessage.
      - If nothing usable is found, returns a default
        new AssistantMessage("Empty response from model for unknown reason").
  - Non‑streaming path:
      - Replace direct use of response.getResult().getOutput() with extractAssistantMessage(response) when selecting the final model response.
  - Streaming path (reasoning logs):
      - Replace direct use of chatResponse.getResult().getOutput() with extractAssistantMessage(chatResponse) so streaming logs reflect the selected assistant message (including tool calls) even if it
        is not in the first generation.

  #### 2. NodeExecutor (graph core)

  File:
  spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/executor/NodeExecutor.java

  - In getEmbedFlux(...), which processes embedded Flux<ChatResponse>:
      - Add a helper:

        private AssistantMessage extractAssistantMessage(ChatResponse response)

        with the same preference rules as in AgentLlmNode, but returning null when no usable AssistantMessage exists (e.g. usage‑only chunks).
      - Update the filtering logic:
          - Previously: keep all ChatResponse objects where response.getResult() != null.
          - Now: only keep ChatResponse objects where extractAssistantMessage(response) != null.
            This naturally filters out usage‑only or non‑assistant chunks before mapping.
      - Update the mapping logic:
          - Replace all direct response.getResult().getOutput() accesses with extractAssistantMessage(response).
          - Maintain lastChatResponseRef as a normalized ChatResponse whose single generation’s output is the selected AssistantMessage.
          - For each new chunk:
              - If there is no previous response:
                  - Stream the current AssistantMessage and normalize lastChatResponseRef.
              - If the current assistant message has tool calls:
                  - Stream that message and replace lastChatResponseRef with a normalized response containing that message.
              - Otherwise:
                  - Concatenate lastMessage.text and currentMessage.text into a new AssistantMessage, update lastChatResponseRef with an aggregated ChatResponse, and stream only the current chunk
                    message.
      - Completion logic:
          - Previously: used lastChatResponseRef.get().getResult().getOutput() directly.
          - Now: uses extractAssistantMessage(lastChatResponseRef.get()) as the final message and stores that in the completion result map (under the key and under messages when applicable).

  #### 3. GraphFluxGenerator (graph core)

  File:
  spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/streaming/GraphFluxGenerator.java

  - In GraphFluxGenerator.Builder:
      - Add a helper:

        private AssistantMessage extractAssistantMessage(ChatResponse response)

        with the same “prefer tool calls, fallback to last assistant” semantics, returning null when no assistant message exists.
      - Update buildInternal(Flux<ChatResponse> flux):
          - The mergeMessage function now:
              - Uses extractAssistantMessage(response) as currentMessage.
              - If currentMessage == null, treat the chunk as usage‑only and keep the previously aggregated ChatResponse.
              - If it is the first chunk, normalize the response into one generation with currentMessage.
              - If it is not the first chunk:
                  - If currentMessage.hasToolCalls(), override the aggregated response with the current message.
                  - Otherwise, concatenate lastMessage.text and currentMessage.text into a new AssistantMessage, build a new aggregated ChatResponse, and return it.
          - The chunk text function passed to GraphFlux.of is updated to:

            response -> {
                AssistantMessage message = extractAssistantMessage(response);
                return message != null ? message.getText() : null;
            }

  This brings GraphFlux‑based streaming behavior in line with the agent/graph streaming behavior.

  ### Rationale

  - Multi‑generation correctness:
    ChatResponse can legitimately have multiple generations, and the first one is not guaranteed to be the one with tool calls. Choosing the right generation is critical for correct agent behavior.
  - Tool‑call friendliness:
    Agents and tools should see the generation that actually contains tool calls. Prioritizing AssistantMessage.hasToolCalls() makes this explicit in both non‑streaming and streaming paths.
  - Streaming robustness:
    Usage‑only chunks and null outputs are a reality with some providers. Filtering these out early and avoiding raw getResult().getOutput() dereferences prevents NullPointerException and subtle bugs.
  - Consistency:
    Non‑streaming and streaming code paths now share a very similar assistant‑message selection strategy, which makes behavior more predictable and easier to reason about.

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
+++ b/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
@@ -295,7 +295,8 @@ public class AgentLlmNode implements NodeActionWithConfig {
 						continue;
 					}
 					Message output = generation.getOutput();
-					if (output instanceof AssistantMessage assistantMessage) {
+					if (output instanceof AssistantMessage) {
+						AssistantMessage assistantMessage = (AssistantMessage) output;
 						if (assistantMessage.hasToolCalls()) {
 							// Prefer the first message that contains tool calls
 							return assistantMessage;
@@ -318,8 +319,11 @@ public class AgentLlmNode implements NodeActionWithConfig {
 		}
 
 		// Fallback: use the first result if available and of the correct type
-		if (response.getResult() != null && response.getResult().getOutput() instanceof AssistantMessage assistant) {
-			return assistant;
+		if (response.getResult() != null) {
+			Message output = response.getResult().getOutput();
+			if (output instanceof AssistantMessage) {
+				return (AssistantMessage) output;
+			}
 		}
 
 		return defaultMessage;
@@ -362,19 +366,21 @@ public class AgentLlmNode implements NodeActionWithConfig {
 
 		for (int i = messages.size() - 1; i >= 0; i--) {
 			Message message = messages.get(i);
-			if (message instanceof UserMessage userMessage) {
+			if (message instanceof UserMessage) {
+				UserMessage userMessage = (UserMessage) message;
 				// Check if outputSchema is already present to avoid duplication
 				if (!userMessage.getText().contains(outputSchema)) {
 					messages.set(i, userMessage.mutate().text(userMessage.getText() + System.lineSeparator() + outputSchema).build());
 				}
 				break;
 			}
-			if (message instanceof AgentInstructionMessage templatedUserMessage) {
-                String newOutputSchema = outputSchema.replace("{", "\\{").replace("}", "\\}");
-                // Check if outputSchema is already present to avoid duplication
-                if (!templatedUserMessage.getText().contains(newOutputSchema)) {
-                	messages.set(i, templatedUserMessage.mutate().text(templatedUserMessage.getText() + System.lineSeparator() + newOutputSchema).build());
-                }
+			if (message instanceof AgentInstructionMessage) {
+				AgentInstructionMessage templatedUserMessage = (AgentInstructionMessage) message;
+				String newOutputSchema = outputSchema.replace("{", "\\{").replace("}", "\\}");
+				// Check if outputSchema is already present to avoid duplication
+				if (!templatedUserMessage.getText().contains(newOutputSchema)) {
+					messages.set(i, templatedUserMessage.mutate().text(templatedUserMessage.getText() + System.lineSeparator() + newOutputSchema).build());
+				}
 				break;
 			}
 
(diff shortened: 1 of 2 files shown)


chickenlj:
Relate to https://github.com/alibaba/spring-ai-alibaba/pull/3138

disaster1-tesk:
Please deal with the conflict

zengchao304:
It looks like the issue has been fixed, close this PR

PR #3340, opened 2025-12-06
fix: memory example get store by right key

Weilong-Qin:
memory example run fail because get runnableConfig by wrong key.

NONE

### Describe how you did it
Get runnableConfig by right key `_AGENT_CONFIG_`


yuluo-yx (commented):
LGTM thx

yuluo-yx (approved):

PR #3383, opened 2025-12-10
fix(graph): NacosMcpGatewayToolCallback toolContext lose

codezkk:
想在基于nacos 为配置中心的 NacosMcpGatewayToolCallback 中拿到类似FunctionToolCallback 的ToolContext，但是发现AgentToolNode 中丢失了

### Describe how you did it
FunctionToolCallback 和MethodToolCallback 以及其他情况合并，个人理解应该不需要特殊处理

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentToolNode.java
+++ b/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentToolNode.java
@@ -212,14 +212,10 @@ public class AgentToolNode implements NodeActionWithConfig {
 			String result;
 			try {
 				// Handle FunctionToolCallback and MethodToolCallback, which support passing state and config in ToolContext.
-				if (toolCallback instanceof FunctionToolCallback<?, ?> || toolCallback instanceof MethodToolCallback) {
-					Map<String, Object> toolContextMap = new HashMap<>(toolContext);
-					toolContextMap.putAll(Map.of(AGENT_STATE_CONTEXT_KEY, state, AGENT_CONFIG_CONTEXT_KEY, config, AGENT_STATE_FOR_UPDATE_CONTEXT_KEY, extraStateFromToolCall));
-					result = toolCallback.call(req.getArguments(), new ToolContext(toolContextMap));
-				} else {
-					// FIXME, currently MCP Tool does not support State and RunnableConfig transmission in ToolContext.
-					result = toolCallback.call(req.getArguments(), new ToolContext(toolContext));
-				}
+                Map<String, Object> toolContextMap = new HashMap<>(toolContext);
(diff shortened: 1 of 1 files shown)


codezkk:
@chickenlj  有空瞅瞅

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentToolNode.java
```diff
@@ -212,14 +212,10 @@ private ToolCallResponse executeToolCallWithInterceptors(
...
-					toolContextMap.putAll(Map.of(AGENT_STATE_CONTEXT_KEY, state, AGENT_CONFIG_CONTEXT_KEY, config, AGENT_STATE_FOR_UPDATE_CONTEXT_KEY, extraStateFromToolCall));
-					result = toolCallback.call(req.getArguments(), new ToolContext(toolContextMap));
-				} else {
-					// FIXME, currently MCP Tool does not support State and RunnableConfig transmission in ToolContext.
-					result = toolCallback.call(req.getArguments(), new ToolContext(toolContext));
-				}
+                Map<String, Object> toolContextMap = new HashMap<>(toolContext);
+                toolContextMap.putAll(Map.of(AGENT_STATE_CONTEXT_KEY, state, AGENT_CONFIG_CONTEXT_KEY, config, AGENT_STATE_FOR_UPDATE_CONTEXT_KEY, extraStateFromToolCall));
+                result = toolCallback.call(req.getArguments(), new ToolContext(toolContextMap));
+
```
chickenlj, line 218: We should exclude 'MCP Tool' here. ToolCallback of MCP type cannot properly handle `RunnableConfig` context map properly by default.

codezkk: Therefore, the toolContext here should not only be a static configuration. For example, when executing a toolcall on the MCP server, developers may want to generate some dynamic headers based on the current user's request and pass them through to the MCP server？

codezkk: Additionally, are there plans to support SPI extensions for implementations similar to NacosMcpGatewayToolCallback, which uses Nacos as the configuration center?

chickenlj: > Context here should not only be a static configuration. For example, when executing a toolcall on the MCP server, developers may want to generate some dynamic headers based on the current user's request and pass them through to the MCP server？

Yes, that's true, you can pass a map to MCPTool. The problem with this specific case is that you have a special key mapping to a value of `RunnableConfig` type. MCPTool will also try to serialize and send it to the mcp server whichi will cause exception.

PR #3758, opened 2025-12-18
feat: add mutate api for ReactAgent

yuluo-yx:
refer: 

- https://docs.spring.io/spring-ai/reference/api/chatclient.html


spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
```diff
@@ -75,6 +76,10 @@ public class AgentLlmNode implements NodeActionWithConfig {
 
 	private ChatClient chatClient;
 
+	private ChatModel chatModel;
```
yuluo-yx, line 79: any ideas?

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
```diff
@@ -75,11 +76,15 @@ public class AgentLlmNode implements NodeActionWithConfig {
...
+
+	private ChatOptions chatOptions;
```
chickenlj, line 81: ChatOptions is not used in any place? What's the purpose of adding this new filed?

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
chickenlj: Ideas for what thing? Could you please explain?

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/Builder.java
```diff
@@ -113,6 +113,36 @@ public abstract class Builder {
...
+		this.compileConfig = agent.getCompileConfig();
+		this.inputSchema = agent.getInputSchema();
+		this.inputType = agent.getInputType();
+		this.outputSchema = agent.getOutputSchema();
+		this.outputType = agent.getOutputType();
+
+		if (this.compileConfig != null && this.compileConfig.checkpointSaver().isPresent()) {
+			this.saver = this.compileConfig.checkpointSaver().get();
+		}
+	}
```
chickenlj, line 144: I'm a bit worried that this is not a `full copy` and `change` action as people expected. Some fields set in the original Builder were used for building AgentLlmNode or AgentToolNode, so they are not all stored in ReactAgent instance (this can be found in the `Builder.build()` method). Simply building a new Builder by making a copy of ReactAgent might not get a new ReactAgent that keeps the same behaviour as the previous one.

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
yuluo-yx: ReactAgent directly uses chatModel 🤔 

It may not be useful at this stage; if it's not needed, I can delete it.

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
yuluo-yx: same as chatModels

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/Builder.java
yuluo-yx: got it, ReactAgent is a stateful instance. If want to get a new ReactAgent instance, might need to give up something. right?

yuluo-yx: I think we can reassess the feasibility of mutate api; if it's not necessary, we should just close this PR.

chickenlj: I think it's good to have `mutable` method. 

Considering the way Builder.build() works -- some of the properties are stored in AgentToolNode and AgentModelNode. To make a ReactAgent copy means also need to make a copy of the AgentToolNode and AgentModelNode attributes of ReactAgent.

I'll close for now, please reopen when ready.

PR #3826, opened 2025-12-19
fix(autoconf): NacosAgentConfig load duplicate key error

codezkk:

When a configuration file repeatedly configures a certain key, for example:
1. In the application.yml configuration, set spring.ai.alibaba.agent.proxy.nacos.namespace to dev (this is written this way because there may be many test environments)
2. In the application-prod.yml configuration, set the value of spring.ai.alibaba.agent.proxy.nacos.namespace to "prod"
At this time, the toMap method of NacosAgentConfig will directly report an error due to duplicate keys, instead of performing an overwrite operation according to Spring's specifications
### Describe how you did it
Let Spring handle priority automatically and remove duplicates

### Describe how to verify it
its easy to see


codezkk:
hello, here I come again, take a look when you have time ,thank you! @chickenlj

chickenlj:
> hello, here I come again, take a look when you have time ,thank you! @chickenlj

Cool! Really appreciate the contribution. I had been busy last week, sorry for the late response.

Issue #4094, opened 2026-01-20
[BUG] ReactAgent添加tools时，如果不设置chatOptions会导致发送给LLM的tools丢失


ADone1020:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

<img width="1467" height="492" alt="Image" src="https://github.com/user-attachments/assets/3bd5dba1-6121-4a17-b35d-9d154dcf3463" />

### Expected Behavior

ReactAgent的chatOptions和tools应该是独立的，不设置chatOptions，tools也应该能正常发送给LLM

### Steps To Reproduce

<img width="1593" height="771" alt="Image" src="https://github.com/user-attachments/assets/c5c282f4-39a5-419f-a374-ccaf0bb984d3" />

<img width="1824" height="990" alt="Image" src="https://github.com/user-attachments/assets/26dbe621-fb33-496c-872c-7d4e611c0fc1" />

<img width="1512" height="612" alt="Image" src="https://github.com/user-attachments/assets/765ed4a2-b2b4-4b95-b608-35b9a04fd82a" />

<img width="1857" height="813" alt="Image" src="https://github.com/user-attachments/assets/54061cb1-12e8-4fe2-bf36-794a74d78463" />

<img width="1782" height="390" alt="Image" src="https://github.com/user-attachments/assets/177e09e1-f2e9-484a-b726-db47f047fd7e" />

### Environment

```markdown
Spring AI Alibaba version(s):1.1.2.0-SNAPSHOT
```

X-qinghai:
Hi there, I noticed this issue and I'm interested in looking into it. However, I'm not entirely sure if this requires a fix or if it's intended behavior. Could you please confirm if this needs to be addressed?

I have a preliminary direction and a rough idea of where the issue might be. If it does need fixing, I'd like to give it a try. Thanks!

taskgirl123:
我使用的ReactAgent搭配OpenAi。
我调整了一天的系统提示词，无论怎么设置都不会调用Tools。
后来做了单元测试才发现了问题，如果在yaml中配置输出格式：
          response_format:
            type: json_object
则tools不会被调用，如果把这个配置去掉，则tools会被调用，但是使用mimo-v2.5以及deepseek-v4-flash时，无法按照json格式输出。ReactAgent中已经配置了outputType也不行。

Issue #4049, opened 2026-01-13
[BUG] 基于配置中心的大模型请求丢失tool信息


codezkk:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

基于配置中心构建的ReactAgent，在和大模型交互时候， 由于AgentLlmNode#filterToolCallbacks 这里不会使用toolCallbacks 并且modelRequest不会设置options 综合导致 丢失了tool信息

private List<ToolCallback> filterToolCallbacks(ModelRequest modelRequest) {
		List<ToolCallback> toolCallbacks = new ArrayList<>();
		if (modelRequest == null) {
			toolCallbacks.addAll(this.toolCallbacks);
		} else if (modelRequest.getOptions() != null && modelRequest.getOptions().getToolCallbacks() != null) {
			toolCallbacks.addAll(modelRequest.getOptions().getToolCallbacks());
		}

		if (modelRequest == null || modelRequest.getTools() == null || modelRequest.getTools().isEmpty()) {
			return toolCallbacks;
		}

		List<String> requestedTools = modelRequest.getTools();
		if (requestedTools == null || requestedTools.isEmpty()) {
			return toolCallbacks;
		}
		return new ArrayList<>(toolCallbacks.stream()
				.filter(callback -> requestedTools.contains(callback.getToolDefinition().name()))
				.toList());
	}

### Environment

```markdown
Spring AI Alibaba version(s): main分支最新代码
```

PR #4112, opened 2026-01-21
fix(agent): correct tool callbacks filtering logic in AgentLlmNode

FatOnionLee:

Fix the filterToolCallbacks method in AgentLlmNode to correctly fallback to instance-level tool callbacks when model request options don't provide any. Previously, when modelRequest.getOptions().getToolCallbacks() was null, the method would not properly use this.toolCallbacks as fallback, potentially resulting in missing tool callbacks.

image

When modelRequest != null && (modelRequest.getOptions() == null || modelRequest.getOptions().getToolCallbacks() == null) && (requestedTools != null || !requestedTools.isEmpty()))，this method will return empty list. Therefore, the upper-level method will not receive any tool.

https://github.com/alibaba/spring-ai-alibaba/issues/4049
https://github.com/alibaba/spring-ai-alibaba/issues/4094

### Describe how you did it

Refactored the conditional logic in filterToolCallbacks method:                                                                 
- Return early with this.toolCallbacks when modelRequest is null                                                                
- Use modelRequest.getOptions().getToolCallbacks() if available                                                                 
- Otherwise, fallback to this.toolCallbacks when options callbacks are null                                                     
- Then apply the tools filtering logic on the collected callback

### Describe how to verify it

Add some unit tests and and ensure they pass
mvn -pl :spring-ai-alibaba-agent-framework -am -Dtest=AgentLlmNodeFilterToolCallbacksTest test

image
image

### Special notes for reviews

NONE

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
+++ b/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/node/AgentLlmNode.java
@@ -447,14 +447,12 @@ public class AgentLlmNode implements NodeActionWithConfig {
 		List<ToolCallback> toolCallbacks = new ArrayList<>();
 		if (modelRequest == null) {
 			toolCallbacks.addAll(this.toolCallbacks);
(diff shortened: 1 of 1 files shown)


FatOnionLee:
Hi @chickenlj , could you please take a look at this PR when you have a moment? It corrects tool callbacks filtering logic in AgentLlmNode and I’d appreciate your feedback.

PR #4139, opened 2026-01-26
fix(graph): add tool returnDirect support

chickenlj:
related to https://github.com/alibaba/spring-ai-alibaba/pull/4075

## Overview

Implements the `returnDirect` feature for tools, allowing tools to return results directly to users without model reprocessing, reducing model calls and improving response speed.

## Core Capabilities

- **Direct Return**: When `returnDirect=true`, tool results are returned directly as final output
- **Auto Termination**: Automatically jumps to END node when `returnDirect` is detected, skipping model processing
- **Multiple Response Support**: Single response returned directly, multiple responses automatically formatted as JSON array

### AgentToolNode
- When all tools have `returnDirect=true`, sets `finishReason` marker in `ToolResponseMessage.metadata`
- Supports both initial execution and continuation execution scenarios

### FinishReasonCheckMessagesModelHook
- **Position**: `beforeModel` hook with highest priority (`Integer.MIN_VALUE`)
- **Detection**: Checks if last message is `ToolResponseMessage` and contains `finishReason`
- **Processing**: Generates `AssistantMessage` and uses `JumpTo.end` to jump to END node

## Usage Example

```java
@Tool(name = "getWeather", returnDirect = true)
public String getWeather(String city) {
    return city + " weather is nice";
}
```

## Testing

- ✅ Single/multiple `returnDirect` tools
- ✅ Normal tool compatibility
- ✅ Hook execution order verification

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/test/java/com/alibaba/cloud/ai/graph/agent/FlowAgentHookTest.java
+++ b/spring-ai-alibaba-agent-framework/src/test/java/com/alibaba/cloud/ai/graph/agent/FlowAgentHookTest.java
@@ -26,6 +26,8 @@ import com.alibaba.cloud.ai.graph.agent.hook.ModelHook;
 import com.alibaba.cloud.ai.graph.agent.utils.HookFactory;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
 import org.springframework.ai.chat.model.ChatModel;
 
 import java.util.List;
@@ -39,6 +41,7 @@ import static org.junit.jupiter.api.Assertions.*;
  * @author haojun.phj (Jackie)
  * @since 2025/01/13
  */
+@EnabledIfEnvironmentVariable(named = "AI_DASHSCOPE_API_KEY", matches = ".+")
 public class FlowAgentHookTest {
 
 	private ChatModel chatModel;


nfzsh:
Check ReactAgent::makeToolsToModelEdge?

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/hook/returndirect/ReturnDirectModelHook.java
```diff
@@ -0,0 +1,177 @@
...
+		} else {
+			// If there are multiple responses, generate a JSON array
+			StringBuilder jsonArray = new StringBuilder("[");
+			for (int i = 0; i < responses.size(); i++) {
+				if (i > 0) {
+					jsonArray.append(",");
+				}
+				String responseData = responses.get(i).responseData();
+				// If responseData is already in JSON format, add it directly; otherwise add it as a string
+				if (responseData.trim().startsWith("{") || responseData.trim().startsWith("[")) {
```
Aias00: `responseData.trim()` can throw NPE if `responseData` is null. Add a null check before calling `trim()`.

chickenlj:
> Check ReactAgent::makeToolsToModelEdge?

`ReturnDirectModelHook` is designed to be the first hook to execute before `Model Node`. It will check `returnDirect` and jump to END if necessary.

chickenlj:
I will merge this pull request now so that some volunteer users can help with testing. 

For any concerns, please feel free to leave more comments here.

PR #4159, opened 2026-01-30
fix: the issue where rmq docker could not be started on MAC

tfgzs:
Fixed the issue where rmq docker could not be started on MAC

https://github.com/alibaba/spring-ai-alibaba/issues/4151


chickenlj:
Have you tried setting GID and UID first before running docker-compose?

Issue #4195, opened 2026-02-04
[BUG] Jackson version conflict: SAA should either remove explicit Jackson version or align with Spring AI


RainYuY:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When I simply import the following two dependencies:

```
<dependency>
    <groupId>com.alibaba.cloud.ai</groupId>
    <artifactId>spring-ai-alibaba-agent-framework</artifactId>
    <version>1.1.2.0</version>
</dependency>

<dependency>
    <groupId>org.springframework.ai</groupId>
    <artifactId>spring-ai-starter-model-zhipuai</artifactId>
</dependency>
```
it triggers the following exception:

> Exception in thread "main" java.lang.NoSuchMethodError: 'com.fasterxml.jackson.annotation.OptBoolean com.fasterxml.jackson.annotation.JsonProperty.isRequired()'

It appears that SAA (spring-ai-alibaba-agent-framework) relies on Jackson 2.19.4 but does not include Jackson Annotations in its dependencies. As a result, the annotations still use version 2.17.0 imported by Spring AI.
SAA should either explicitly import Jackson Annotations or delegate the version management of Jackson to Spring AI.

### Environment

```markdown
Spring AI Alibaba version(s):
```

PR #4203, opened 2026-02-05
fix:Add exclusions for jackson-annotations dependency

RainYuY:

fix https://github.com/alibaba/spring-ai-alibaba/issues/4195


RainYuY:
@chickenlj PTAL

PR #4376, opened 2026-03-05
fix(graph): prevent resume metadata leak to non-resumed compiled sub-…

wtJoker:
  When a parent graph has multiple compiled sub-graphs and the first sub-graph interrupts then resumes, the second sub-graph fails
  with `Resume request without a valid checkpoint!`.
  Root cause: `SubCompiledGraphNodeAction.apply()` creates the sub-graph's `RunnableConfig` via `RunnableConfig.builder(config)`,
  which copies all parent metadata including `HUMAN_FEEDBACK_METADATA_KEY`. Non-resumed sub-graphs inherit this key, causing their
  `GraphRunnerContext` to incorrectly enter resume mode.
  NONE
  ### Describe how you did it
  Added a check after constructing `subGraphRunnableConfig`: if this sub-graph is not being resumed (`resumeSubgraph == false`),
  remove the inherited `HUMAN_FEEDBACK_METADATA_KEY` from its metadata.
  ### Describe how to verify it
  1. Create a parent graph with two compiled sub-graphs (A and B), where A has `interruptAfter`
  2. Execute the parent graph — A interrupts
  3. Resume — A completes, then B executes
  4. Without fix: B fails with "Resume request without a valid checkpoint!"
  5. With fix: B executes normally, parent reaches END
  ### Special notes for reviews
  The fix is 3 lines of logic. The per-sub-graph resume mechanism via `resumeSubGraphId` metadata is correct; the issue is only that
  the global resume signal (`HUMAN_FEEDBACK_METADATA_KEY`) leaks to sub-graphs that don't need it.

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/internal/node/SubCompiledGraphNodeAction.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/internal/node/SubCompiledGraphNodeAction.java
@@ -93,6 +93,12 @@ public record SubCompiledGraphNodeAction(String nodeId, CompileConfig parentComp
 			}
 		}
 
+		// Prevent non-resumed sub-graphs from inheriting parent's resume metadata
+		if (!resumeSubgraph) {
+			subGraphRunnableConfig.metadata()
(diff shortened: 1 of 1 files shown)


robocanic:
@robocanic

PR #4385, opened 2026-03-08
feat(graph): add langfuse tracing support for graph and agent flows

xiaohai-78:

This PR wires `ObservationRegistry` through the agent and graph layers so that
Spring AI chat/model observations and graph execution observations are attached
to the same trace. It also adds a Langfuse-based verification path in `spring-ai-alibaba-studio`.

We need this because previously the registry set on the agent builder only affected
the Spring AI model layer, while graph execution observations were not consistently
connected to the same registry.

NONE

### Describe how you did it

- Added `GraphObservationSupport` to centralize graph observation enhancement
- Updated agent builders to propagate `ObservationRegistry` into graph `CompileConfig`
- Reused the same enhancement logic in graph observation auto-configuration
- Fixed `CompileConfig` copy behavior to avoid shared listener state
- Added Langfuse profile config and verification tests in `spring-ai-alibaba-studio`

### Describe how to verify it

- Run:
  - `./mvnw -pl spring-boot-starters/spring-ai-alibaba-starter-graph-observation -Dtest=GraphObservationAutoConfigurationTest test`
  - `./mvnw -pl spring-ai-alibaba-studio -Dtest=LangfuseProfileContextTest test`
- With `AI_DASHSCOPE_API_KEY`, `LANGFUSE_ENABLED=true`, and `LANGFUSE_OTEL_AUTH` set, run:
  - `./mvnw -pl spring-ai-alibaba-studio -Dtest=LangfuseAgentStudioLiveTest test`
- Check in Langfuse that the trace includes HTTP, agent/graph, and Spring AI model spans

### Special notes for reviews

- Langfuse verification is test-focused; core behavior changes are in agent-framework, graph-core, and starter graph observation
- The studio live test suppresses known observation context propagation noise in test logs

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/observation/GraphObservationSupport.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/observation/GraphObservationSupport.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.cloud.ai.graph.observation;
+
+import com.alibaba.cloud.ai.graph.CompileConfig;
+import com.alibaba.cloud.ai.graph.GraphLifecycleListener;
+import io.micrometer.observation.ObservationRegistry;
+
+import java.util.Objects;
+
+/**
+ * Utilities for attaching graph observation support to compile configurations.
+ */
+public final class GraphObservationSupport {
+
+	private GraphObservationSupport() {
+	}
+
+	/**
+	 * Enhances a compile configuration with graph observation support.
+	 * @param compileConfig the base compile configuration
+	 * @param observationRegistry the observation registry to apply
+	 * @return the enhanced compile configuration
+	 */
+	public static CompileConfig enhance(CompileConfig compileConfig, ObservationRegistry observationRegistry) {
+		CompileConfig baseConfig = compileConfig != null ? compileConfig : CompileConfig.builder().build();
+		ObservationRegistry effectiveRegistry = observationRegistry != null ? observationRegistry
+				: baseConfig.observationRegistry();
+
+		if (effectiveRegistry == null || effectiveRegistry == ObservationRegistry.NOOP) {
+			return baseConfig;
+		}
+
+		GraphObservationLifecycleListener existingListener = findGraphObservationLifecycleListener(baseConfig);
+		boolean hasLifecycleListener = existingListener != null;
+		boolean sameRegistry = baseConfig.observationRegistry() == effectiveRegistry;
+		boolean listenerMatchesRegistry = existingListener == null
+				|| Objects.equals(existingListener.observationRegistry(), effectiveRegistry);
+
+		if (sameRegistry && hasLifecycleListener && listenerMatchesRegistry) {
+			return baseConfig;
+		}
+
+		CompileConfig.Builder builder = CompileConfig.builder(baseConfig).observationRegistry(effectiveRegistry);
+		if (hasLifecycleListener && !listenerMatchesRegistry) {
+			builder = CompileConfig.builder(baseConfig)
+					.observationRegistry(effectiveRegistry);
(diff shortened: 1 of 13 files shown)


robocanic:
@disaster1-tesk

KennanChan:
Any updates about this PR?
I am integrating Langfuse and the trace is incomplete.

spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/observation/GraphObservationSupport.java
```diff
@@ -0,0 +1,110 @@
...
+		boolean listenerMatchesRegistry = existingListener == null
+				|| Objects.equals(existingListener.observationRegistry(), effectiveRegistry);
+
+		if (sameRegistry && hasLifecycleListener && listenerMatchesRegistry) {
+			return baseConfig;
+		}
+
+		CompileConfig.Builder builder = CompileConfig.builder(baseConfig).observationRegistry(effectiveRegistry);
+		if (hasLifecycleListener && !listenerMatchesRegistry) {
+			builder = CompileConfig.builder(baseConfig)
```
oss-sentinel-ai, line 59: This replacement path starts with CompileConfig.builder(baseConfig), which already copies every listener. The following filter only appends non-graph listeners and never removes the existing GraphObservationLifecycleListener, so a different builder registry leaves the old listener in place, duplicates every other listener, and adds a second graph listener. Build from a config without the old listener (or add a listener-replacement API) before adding the new listener.

spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/observation/GraphObservationSupport.java
```diff
@@ -0,0 +1,110 @@
...
+					.observationRegistry(effectiveRegistry);
+
+			baseConfig.lifecycleListeners()
+					.stream()
+					.filter(listener -> !(listener instanceof GraphObservationLifecycleListener))
+					.forEach(builder::withLifecycleListener);
+		}
+
+		if (!hasLifecycleListener || !listenerMatchesRegistry) {
+			builder.withLifecycleListener(new GraphObservationLifecycleListener(effectiveRegistry));
```
oss-sentinel-ai, line 69: Attaching this listener exposes graph executions that end via interruption, max-iteration/stop, embedded return, or subscriber cancellation: GraphObservationLifecycleListener only stops/unregisters on END or ERROR, while MainGraphExecutor returns done directly for those terminal paths. The static context entry and started observation can therefore leak. Add an abort/cancel terminal callback (or a doFinally-style cleanup) that stops and unregisters the graph observation exactly once.

oss-sentinel-ai (changes requested):
### Summary
Adds `GraphObservationSupport` to wire an `ObservationRegistry` and `GraphObservationLifecycleListener` into `CompileConfig` from the agent builders (`Builder`, `FlowAgentBuilder`) and the graph-observation auto-configuration, fixes `CompileConfig` copy semantics, and adds an opt-in Langfuse profile with docs/tests in Studio. The direction is right and the common same-registry path is idempotent, but there are two observation-lifecycle correctness problems in graph-core that should be addressed before merging.

### Findings
- **[Warning]** `GraphObservationSupport.java:59` — registry-replacement path duplicates lifecycle listeners instead of replacing the graph listener (old listener kept, other listeners duplicated).
- **[Warning]** `GraphObservationSupport.java:69` — graph observations can leak on terminal paths that bypass END/ERROR (interrupt, max-iteration/stop, embedded return, subscriber cancellation); no abort/cancel cleanup callback.
- **[Warning]** `LangfuseAgentStudioLiveTest.java:587` — a non-blank SSE response does not verify tracing; assert graph and chat/model spans share one trace via a recording handler/exporter, and add non-live tests.

### Suggestions
1. In `GraphObservationSupport.enhance`, when switching registries, build from a config with the old `GraphObservationLifecycleListener` removed (or add a listener-replacement API) before attaching the new one, so listeners are replaced rather than accumulated.
2. Add a terminal abort/cancel callback (or `doFinally`-style cleanup) that stops and unregisters the graph observation exactly once for the interrupt/stop/max-iteration/embedded-return/cancellation paths in `MainGraphExecutor`.
3. Note that pre-built nested sub-agents do not inherit the flow builder's registry (`FlowGraphBuildingStrategy` embeds them as-is) — worth documenting that each child agent needing chat/model observations must configure its own registry.
4. Add focused unit tests for same-registry/different-registry enhancement, supplied `CompileConfig` merge semantics, and interrupt/error/cancellation cleanup; verify span parentage on async parallel branches.

---

### Additional notes (not anchored to a changed line)
- **[WARNING]** `spring-ai-alibaba-studio/src/test/java/com/alibaba/cloud/ai/langfuse/LangfuseAgentStudioLiveTest.java:587` — A nonblank SSE response does not verify tracing: this passes if GraphObservationSupport is never applied, graph/chat spans are separate, or OTLP export fails. Record observations with a test handler/exporter and assert the graph and chat/model spans share one trace (and add a non-live test for the supplied-CompileConfig and different-registry cases). _(line outside diff)_

PR #4395, opened 2026-03-10
fix(agent): Fixed ShellSessionManager's indefinite waiting for a marker to appear.

codezkk:
ShellSessionManager uses a marker command (such as `printf '__LC_SHELL_DONE__ %s\n' $?`) to detect when a command has finished executing and extracts the exit code. If the output of a command (such as `curl -s`, `printf`, or some scripts) does not end with a newline character, the shell marker is appended directly to the same line, for example:

{"status":"ok"}__LC_SHELL_DONE__... 0

Because the original code uses `line.startsWith(marker)` (to check if the line starts with a marker),

it cannot find the marker (because the marker is in the middle of the line, not at the beginning). The tool will wait indefinitely for the marker to appear until a timeout occurs, causing the program to hang and restart the session.

### Describe how you did it
ShellSessionManager.java has been updated:

1. Replaced line.startsWith(marker) with line.contains(marker), ensuring correct recognition even when the marker follows command output.

2. Correctly splits lines, extracts command output preceding the marker, and extracts exit codes.

### Describe how to verify it
see ShellSessionManagerTest

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/tools/ShellSessionManager.java
+++ b/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/tools/ShellSessionManager.java
@@ -354,83 +354,105 @@ public class ShellSessionManager {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


robocanic:
@yuluo-yx

PR #4429, opened 2026-03-17
fix: properly check returnDirect in makeToolsToModelEdge to fix premature workflow termination

### LINKED ISSUE #4427: [BUG] ReactAgent returnDirect错误

### Current Behavior

	private EdgeAction makeToolsToModelEdge(String modelDestination, String endDestination) {
		return state -> {
			// 1. Extract last AI message and corresponding tool messages
			ToolResponseMessage toolResponseMessage = fetchLastToolResponseMessage(state);
			// 2. Exit condition: All executed tools have return_direct=True
			if (toolResponseMessage != null && !toolResponseMessage.getResponses().isEmpty()) {
				boolean allReturnDirect = toolResponseMessage.getResponses().stream().allMatch(toolResponse -> {
					String toolName = toolResponse.name();
					return false; // FIXME
				});
				if (allReturnDirect) {
					return endDestination;
				}
			}

			// 3. Default: Continue the loop
			//    Tool execution completed successfully, route back to the model
			//    so it can process the tool results and decide the next action.
			return modelDestination;
		};
	}

我有一个tool设置为returnDirect，第一次调用时，工具内容可以直接返回，第二次再调用（会话保持一致）时，会直接检测历史消息包含了returnDirect，调用end节点.导致不经过大模型节点

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
Spring AI Alibaba version(s):
```

### Debug logs

_No response_

### Anything else?

_No response_

zhexueqi:
## Problem

When a tool has 
eturnDirect=true, the ReactAgent should route directly to the end node after tool execution, bypassing the model. However, the routing logic in makeToolsToModelEdge had a FIXME placeholder that always returned alse, making the 
eturnDirect feature completely non-functional.

Additionally, the issue reporter found a second bug: on the **second call** (with the same session), the agent would incorrectly terminate early because the ToolResponseMessage from the previous turn still had the inish_reason metadata in the message history.

## Root Cause

In ReactAgent.makeToolsToModelEdge:

`java
boolean allReturnDirect = toolResponseMessage.getResponses().stream().allMatch(toolResponse -> {
    String toolName = toolResponse.name();
    return false; // FIXME  ? always false, returnDirect never works
});
`

## Fix

AgentToolNode already correctly sets inish_reason metadata on ToolResponseMessage when all tools have 
eturnDirect=true. The fix reads this metadata from the **last** ToolResponseMessage in state (which etchLastToolResponseMessage already retrieves correctly - it always gets the most recent one, not historical ones).

`java
Map<String, Object> metadata = toolResponseMessage.getMetadata();
if (metadata != null && FINISH_REASON_METADATA_KEY.equals(metadata.get(FINISH_REASON_METADATA_KEY))) {
    return endDestination;
}
`

This correctly handles both issues:
1. 
eturnDirect=true now actually works
2. Historical ToolResponseMessage entries are ignored because etchLastToolResponseMessage only returns the most recent one

Fixes #4427


robocanic:
@yuluo-yx

yuluo-yx:
@zhexueqi ths for your pr, can you provider a simple demo?

btw, i think tests need to be added 👀

PR #4543, opened 2026-04-13
fix: handle checkpoint release Tag and avoid duplicate completion replay

### LINKED ISSUE #4538: [Bug]   Title: handleCompletion 返回 BaseCheckpointSaver.Tag 导致 processGraphResponseFlux 抛出 IllegalArgumentException
### Question

Body:                                                                                                                                                                         
                                                                                                                                                                                
  Bug Description                                                                                                                                                               
                                                                                                                                                                                
  When releaseThread=true and a CheckpointSaver is configured, BaseCheckpointSaver#handleCompletion calls checkpointSaver.release() which returns a BaseCheckpointSaver.Tag     
  object. This Tag is wrapped in GraphResponse.done(tag) and flows back into NodeExecutor#processGraphResponseFlux, where the code at line 556-563 expects a Map but gets a Tag,
   causing:                                                                                                                                                                     
                                                                                                                                                                                
  java.lang.IllegalArgumentException: Node stream must return Map result using Data.done(),                                                                                     
                                                                                                                                                                                
  Root Cause                                                                                                                                                                    
                                                                                                                                                                                
  In NodeExecutor.java:588:                                                                                                                                                     
                  
  return processedFlux                                                                                                                                                          
      .concatWith(updateContextMono.thenMany(                                                                                                                                   
          Flux.defer(() -> mainGraphExecutor.execute(context, resultValue))                                                                                                     
      ));                                                                                                                                                                       
                                                                                                                                                                                
  When mainGraphExecutor.execute() reaches handleCompletion() (via END node or shouldStop), it emits GraphResponse.done(tag) where tag is a BaseCheckpointSaver.Tag. This       
  response flows into the processedFlux pipeline where lastData captures it, and then updateContextMono (line 556-563) tries to cast it to Map<?, ?> and fails.                 
                                                                                                                                                                                
  Steps to Reproduce                                                                                                                                                            
   
  1. Configure a graph with releaseThread(true) and a CheckpointSaver (e.g., MemorySaver)                                                                                       
  2. Use a streaming node in the graph
  3. Run the graph — when execution completes, handleCompletion returns a Tag which flows back to processGraphResponseFlux and causes the error                                 
                                                                                                                                                                                
  Suggested Fix                                                                                                                                                                 
                                                                                                                                                                                
  Option A: In NodeExecutor#processGraphResponseFlux, handle Tag type in the updateContextMono:                                                                                 
                  
  if (nodeResultValue.isPresent()) {                                                                                                                                            
      Object value = nodeResultValue.get();                                                                                                                                     
      if (value instanceof Map<?, ?>) {                                                                                                                                         
          updateState = (Map<String, Object>) value;                                                                                                                            
      }                                                                                                                                                                         
      else if (value instanceof BaseCheckpointSaver.Tag) {                                                                                                                      
          // handleCompletion already finalized state, skip update                                                                                                              
      }                                                                                                                                                                         
      else {                                                                                                                                                                    
          throw new IllegalArgumentException("Node stream must return Map result using Data.done(),");                                                                          
      }                                                                                                                                                                         
  }                                                                                                                                                                             
                                                                                                                                                                                
  Option B: handleCompletion should not emit Tag into the GraphResponse.done() stream. Instead, it should only be returned as the final result to the outermost caller.         
                  
  Environment                                                                                                                                                                   
                  
  - Spring AI Alibaba version: latest main branch                                                                                                                               
  - JDK 17

zshs000:
This PR fixes completion handling for embedded graph streams in graph-core.

When `releaseThread=true` and a `CheckpointSaver` is configured, graph completion may produce
`BaseCheckpointSaver.Tag` instead of a state-update `Map`. In embedded graph stream scenarios, that completion value
could be consumed by `NodeExecutor#processGraphResponseFlux(...)`, which previously assumed the result must always be
a `Map` and threw `IllegalArgumentException`.

This PR also avoids replaying completion twice when the upstream embedded stream already emits a
`GraphResponse.done(...)`.

Fixes #4538

### Describe how you did it
- Updated `NodeExecutor#processGraphResponseFlux(...)` to recognize `BaseCheckpointSaver.Tag` as a non-state
  completion result and skip state merge for that case.
- Updated `NodeExecutor#transformFluxToGraphResponse(...)` to avoid replaying completion when the upstream stream
  already emits a completion `GraphResponse`.
- Added `Issue4538ReproductionTest` to reproduce the embedded `graphResponseStream(...)` scenario with
  `releaseThread(true)` and `MemorySaver`.
- Kept the existing completion behavior for normal graph completion paths and regular streaming flows.

### Describe how to verify it
Run the reproduction test:

`mvn -q -pl spring-ai-alibaba-graph-core -Dtest=Issue4538ReproductionTest test`

Run the full graph-core test suite:

`mvn -q -pl spring-ai-alibaba-graph-core test`

Expected result:
- Issue4538ReproductionTest passes
- spring-ai-alibaba-graph-core test suite passes
- embedded graph stream completion no longer throws
  `IllegalArgumentException: Node stream must return Map result using Data.done()`
- completion is no longer replayed twice for upstream `GraphResponse.done(...)`

### Special notes for reviews
A few notes for reviewers:

- The fix is intentionally minimal and localized to NodeExecutor.
- `BaseCheckpointSaver.Tag` is still preserved as a completion result; it is only skipped during embedded flux state
  merge because it is not a state-update Map.
- The completion replay change only affects upstream streams that already emit GraphResponse completion events. Raw
  streaming inputs such as normal `Flux<String>` / `Flux<ChatResponse>` still follow the existing completion wrapping path.
- I also verified spring-ai-alibaba-graph-core tests pass after this change.


yuluo-yx (commented):
LGTM cc @guanxuc

guanxuc (approved):
LGTM

PR #4555, opened 2026-04-16
fix(graph): clear jump_to state when hook does not specify a target

nfzsh:
                                                                                                                                                      
  Fixes a bug where the agent skips model processing on subsequent turns after a `returnDirect=true` tool execution. The stale `jump_to="end"` state  
  persists across turns via checkpoints and is never cleared, causing the edge router to route directly to END.                                       
                                                                                                                                                      
  Fixes #                                                                                                                                             
                                                                           
  ### Describe how you did it                                                                                                                         
                                                                           
  Always write `jump_to` to the state update map in `BeforeModelAction.apply()` and `AfterModelAction.apply()`, setting it to `null` when no jump     
  target is specified. Previously the field was only written when `getJumpTo()` returned a non-null value, leaving stale state uncleared.
                                                                                                                                                      
  ### Describe how to verify it                                                                                                                       
                                         
  1. Configure an agent with a `returnDirect=true` tool and a `MemorySaver` checkpoint                                                                
  2. First turn: call the agent with a message that triggers the returnDirect tool
  3. Second turn: send a normal chat message (e.g., "hello")                                                                                          
  4. Expected: the model processes the second message normally and returns a response                                                                 
                                                                                                                                                      
  ### Special notes for reviews                                                                                                                       
                                                                                                                                                      
  The change is in `MessagesModelHook.java` — both `BeforeModelAction` and `AfterModelAction` are updated with the same logic. Setting `jump_to` to   
  `null` is equivalent to the key not existing for all downstream consumers (`OverAllState.value()` uses `Optional.ofNullable`), so no other logic is
  affected.


yuluo-yx (commented):
看起来没有做多少修改，没有太大的合并价值对我来说，

另外，我不确定下游是否有对 result["jump_to"] 的 null 检查。

可以在检查下的

nfzsh:
> 看起来没有做多少修改，没有太大的合并价值对我来说，
> 
> 另外，我不确定下游是否有对 result["jump_to"] 的 null 检查。
> 
> 可以在检查下的

问题在于：当触发过一次returnDirect=true function调用后，之后的对话会无法经过LLM直接返回，时间有点久远，大概是因为这个判空处理导致后续只能识别到end，在这手动判断填充值以跳出这个死循环，目前在我这里是原位替代修复使用的

yuluo-yx (commented):
I'm not sure, pls take a look @guanxuc

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/hook/messages/MessagesModelHook.java
```diff
@@ -108,9 +108,10 @@ public CompletableFuture<Map<String, Object>> apply(OverAllState state, Runnable
 					result.put("messages", command.getMessages());
 				}
 			}
-			if (command.getJumpTo() != null) {
-				result.put("jump_to", command.getJumpTo().name());
-			}
+			// Always update jump_to to prevent stale state from persisting across turns.
+			// When the hook doesn't specify a jump target, set to null to clear any
+			// previous value (e.g., from a returnDirect=true execution in a prior turn).
+			result.put("jump_to", command.getJumpTo() != null ? command.getJumpTo().name() : null);
```
oss-sentinel-ai, line 114: No regression test added. A test simulating a returnDirect=true tool execution on turn N (which sets jump_to=end) followed by a turn N+1 where no hook sets jump_to would validate the fix and prevent regressions. Consider adding one that asserts the edge router does not route to END on the subsequent turn.

spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/hook/messages/MessagesModelHook.java
```diff
@@ -142,9 +143,10 @@ public CompletableFuture<Map<String, Object>> apply(OverAllState state, Runnable
...
```
oss-sentinel-ai, line 149: Same fix applied symmetrically to AfterModelAction — correct. Both BeforeModelAction and AfterModelAction now unconditionally write jump_to, ensuring stale state is cleared regardless of which hook phase last set a jump target.

oss-sentinel-ai (commented):
﻿### Summary
This PR fixes a real bug: after a `returnDirect=true` tool execution sets `jump_to="end"`, the value persists through checkpoints and makes the edge router skip model processing on subsequent turns. Making `BeforeModelAction`/`AfterModelAction` always write `jump_to` (null when no target is specified) is a sound fix direction. Verdict: **comment** — the change is correct for `MessagesModelHook`, but the identical conditional pattern in `MessagesAgentHook` is left unaddressed, so the stale-state bug can still leak through agent-level hooks.

Regarding the open question in the existing review thread (downstream null-safety of `jump_to`): verified against current `main` —
- `OverAllState.data` is a plain `HashMap` (accepts null); no `KeyStrategy` is registered for `jump_to`, so default `ReplaceStrategy` applies and stores the null.
- The edge router reads via `state.value("jump_to")` → `Optional.ofNullable(...)` → empty → `resolveJump(null, ...)` returns the default destination, i.e. "no jump". Correct behavior.
- The static `updateState` path strips nulls (`toMapRemovingNulls`), also safe.
- Checkpoint serializers (`JacksonStateSerializer`, `ObjectStreamStateSerializer`, Redis/FileSystem savers) all round-trip null map values safely.

### Findings
- **[Warning]** `MessagesAgentHook.java:94` / `:128` (not in this diff) — `BeforeAgentAction`/`AfterAgentAction` still use the conditional `if (command.getJumpTo() != null)` pattern; a `beforeAgent`/`afterAgent` hook that sets `jump_to` on one turn and omits it on the next reproduces the exact same stale-state bug. Apply the same unconditional-write pattern there (or document why agent-level hooks are unaffected).
- **[Info]** No regression test added — a test simulating a `returnDirect=true` tool call on turn N followed by a normal turn N+1 (asserting the router does not jump to END) would lock in this fix.
- **[Info]** The symmetric change in `AfterModelAction` is correct — stale state is cleared regardless of which hook phase last set a jump target.

### Suggestions
1. Extend the same one-line pattern to `MessagesAgentHook.BeforeAgentAction`/`AfterAgentAction` so all hook phases clear `jump_to` consistently.
2. Add a graph-level regression test for the returnDirect → next-turn scenario described above.

---

### Additional notes (not anchored to a changed line)
- **[WARNING]** `spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/hook/messages/MessagesAgentHook.java:114` — Incomplete fix: MessagesAgentHook.BeforeAgentAction (line 94) and AfterAgentAction (line 128) have the same conditional `if (command.getJumpTo() != null)` pattern. If a beforeAgent/afterAgent hook sets jump_to on one turn and omits it on the next, the same stale-state bug will persist through checkpoints. The unconditional-write pattern should be applied here too, or the author should document why agent-level hooks are not affected. _(line outside diff)_

PR #4586, opened 2026-05-05
feat(autoconf): add message-channel starter for IM integrations

### LINKED ISSUE #4568: [Feature] 希望能支持消息渠道
### Feature Request

希望能跟上python的主流，提供主流消息渠道支持

### Is your feature request related to a problem? Please describe

_No response_

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

i-zoufeng:
Closes #4568

Adds spring-ai-alibaba-starter-message-channel: a unified IM-channel layer that lets Agents talk to DingTalk, WeChat Work, Feishu, Telegram, Slack and similar platforms without each adapter touching the dispatcher or webhook routing.

Core abstractions:
- ChannelMessage / ChannelReply: neutral inbound/outbound models
- MessageChannelAdapter: per-platform contract covering signature verification, payload parsing, sync-reply serialization, active push
- ChannelAdapterRegistry, AgentBindingRegistry, MessageChannelDispatcher: routes inbound messages to the bound Agent and threads channelName:conversationId as RunnableConfig.threadId so existing checkpointers segregate IM conversations
- MessagePublisher: Agent-initiated push entry point
- MessageChannelController: single POST /channel/{name}/callback webhook

Reference DingTalk adapter implements both directions:
- Inbound HMAC-SHA256 verification with replay window and constant-time comparison
- Sync reply via DingTalk msgtype envelope (text + markdown)
- Active push to a configured robot webhook URL

Spring Boot auto-configuration builds one adapter per channels.<name> entry; user-defined adapter beans override built-ins by name.

Includes 18 unit tests (DingTalk adapter + Controller + AutoConfig) and a runnable example under examples/message-channel-dingtalk demonstrating push and inbound wiring.


Aias00 (changes requested):
The overall direction looks good, but I don’t think this is ready to merge yet. There are three blocking issues:

1. The DingTalk reply path is implemented incorrectly.
`MessageChannelController` currently prefers `serializeSyncReply()`, and `DingTalkChannelAdapter` returns the agent reply directly in the webhook HTTP response body.
For DingTalk, replies in this scenario should go through `sessionWebhook` or the corresponding reply API. The current implementation parses `sessionWebhook` in `parseInbound()`, but never actually uses it.
As a result, the service may return HTTP 200 successfully while the user never receives a reply in DingTalk.

2. The proactive push signing configuration is incomplete.
The example documents outbound push as "custom group robot with signing enabled", but `DingTalkChannelAdapter.push()` signs outbound requests using `appSecret`, and there is no separate webhook secret in the configuration.
That means the proactive push flow described by the example cannot actually be configured correctly with the current code.

3. The example project does not build independently.
`examples/message-channel-dingtalk/pom.xml` imports the released `spring-ai-alibaba-bom:1.1.2.2`, but that BOM does not yet include the newly added `spring-ai-alibaba-starter-message-channel`.
Running `./mvnw -f examples/message-channel-dingtalk/pom.xml test -DskipTests` fails immediately because the version for `spring-ai-alibaba-starter-message-channel` is missing.

One non-blocking issue that should also be addressed soon:
- `DingTalkChannelAdapter.parseInbound()` only reads `text.content` and does not branch on `msgtype`. For non-text inbound messages, the agent will likely receive an empty prompt instead of a translated message or an explicit rejection.

Local verification:
- `./mvnw -pl spring-boot-starters/spring-ai-alibaba-starter-message-channel test -DskipITs` passed
- `./mvnw -f examples/message-channel-dingtalk/pom.xml test -DskipTests` failed as described above

I recommend fixing the reply/push model and the example dependency/build issue before merging.

examples/message-channel-dingtalk/src/main/resources/application.yml
```diff
@@ -0,0 +1,51 @@
...
+        options:
+          model: qwen-plus
+    alibaba:
+      message-channel:
+        channels:
+          dingtalk:
+            type: dingtalk
+            bind-agent: customerAgent
+            options:
+              app-secret: ${DINGTALK_APP_SECRET:please-set-DINGTALK_APP_SECRET}
```
oss-sentinel-ai, line 46: This non-empty fallback passes the starter's required-secret validation, so running the example without DINGTALK_APP_SECRET uses a publicly known HMAC key and lets anyone forge callbacks. Default to an empty/unresolved value and fail startup instead.

examples/message-channel-dingtalk/src/main/java/com/alibaba/cloud/ai/examples/dingtalk/DemoPushController.java
```diff
@@ -0,0 +1,53 @@
...
+	private final MessagePublisher publisher;
+
+	public DemoPushController(MessagePublisher publisher) {
+		this.publisher = publisher;
+	}
+
+	/** Browser-friendly: open http://localhost:8080/demo/push?text=hello in any tab. */
+	@GetMapping("/demo/push")
+	public Map<String, String> pushViaQuery(@RequestParam(defaultValue = "Hello from agent") String text) {
+		publisher.pushText("dingtalk", null, null, text);
```
oss-sentinel-ai, line 41: The example tells users to expose this application through a public tunnel, but this unauthenticated endpoint relays arbitrary caller-controlled text to the configured DingTalk webhook. Remove it from the public example or protect it behind a demo/local-only profile.

spring-boot-starters/spring-ai-alibaba-starter-message-channel/src/main/java/com/alibaba/cloud/ai/messagechannel/adapter/dingtalk/DingTalkChannelAdapter.java
```diff
@@ -0,0 +1,216 @@
...
+		return new ChannelMessage(name, senderId, conversationId, text, null, attrs);
+	}
+
+	@Override
+	public Object serializeSyncReply(ChannelMessage message, ChannelReply reply) {
+		return toDingTalkEnvelope(reply);
+	}
+
+	@Override
+	public void push(String userId, String conversationId, ChannelReply reply) {
```
oss-sentinel-ai, line 119: push(userId, conversationId, reply) ignores both destination arguments and always posts to the configured group webhook. This violates the adapter contract and can broadcast a message intended for one user; reject targeted calls until OAuth targeting is implemented, or implement it.

spring-boot-starters/spring-ai-alibaba-starter-message-channel/src/main/java/com/alibaba/cloud/ai/messagechannel/adapter/dingtalk/DingTalkChannelAdapter.java
```diff
@@ -0,0 +1,216 @@
...
+			throw new SecurityException("Timestamp outside the 5-minute replay window");
+		}
+		String expected = sign(timestamp + "\n" + appSecret);
+		if (!constantTimeEquals(expected, sign)) {
+			throw new SecurityException("Signature mismatch");
+		}
+	}
+
+	private String signWebhookCall(long timestamp) {
+		return sign(timestamp + "\n" + appSecret);
```
oss-sentinel-ai, line 183: Outbound custom-robot requests are signed with the inbound enterprise appSecret. The documented custom robot has its own signing secret and no webhook-secret property is available, so signed active push will fail. Configure and use a separate webhook signing secret.

spring-boot-starters/spring-ai-alibaba-starter-message-channel/src/main/java/com/alibaba/cloud/ai/messagechannel/dispatcher/MessageChannelDispatcher.java
```diff
@@ -0,0 +1,95 @@
...
+	private static final Logger log = LoggerFactory.getLogger(MessageChannelDispatcher.class);
+
+	private final AgentBindingRegistry bindings;
+
+	public MessageChannelDispatcher(AgentBindingRegistry bindings) {
+		this.bindings = bindings;
+	}
+
+	public ChannelReply dispatch(ChannelMessage message) {
+		Agent agent = bindings.resolve(message.channelName());
```
oss-sentinel-ai, line 53: Agent binding resolution is outside the try block. A channel without bind-agent, or one bound to a missing/wrong-type bean, escapes as an unhandled webhook 500. Validate bindings at startup or map this to a controlled response.

spring-boot-starters/spring-ai-alibaba-starter-message-channel/src/main/java/com/alibaba/cloud/ai/messagechannel/dispatcher/MessageChannelDispatcher.java
```diff
@@ -0,0 +1,95 @@
...
+				.threadId(threadIdFor(message))
+				.build();
+		try {
+			Optional<OverAllState> result = agent.invoke(message.text(), config);
+			return result.map(this::toReply).orElseGet(() -> ChannelReply.text(""));
+		}
+		catch (Exception e) {
+			log.error("Agent invocation failed for channel={} user={}",
+					message.channelName(), message.userId(), e);
+			return ChannelReply.text("[agent error] " + e.getMessage());
```
oss-sentinel-ai, line 64: The exception message is sent to the external IM user as normal reply text. Return a generic user-safe failure message and retain the exception details only in logs/metrics.

spring-boot-starters/spring-ai-alibaba-starter-message-channel/src/main/java/com/alibaba/cloud/ai/messagechannel/web/MessageChannelController.java
```diff
@@ -0,0 +1,126 @@
...
+			return ResponseEntity.ok(syncBody);
+		}
+
+		try {
+			adapter.push(message.userId(), message.conversationId(), reply);
+		}
+		catch (Exception e) {
+			log.error("Async push failed for channel '{}'", name, e);
+		}
+		return ResponseEntity.ok(Collections.emptyMap());
```
oss-sentinel-ai, line 113: Async push failures are caught and then acknowledged with HTTP 200, so the webhook provider cannot retry and the message is silently lost. Propagate a retryable failure or hand delivery to a durable queue.

spring-boot-starters/spring-ai-alibaba-starter-message-channel/src/main/java/com/alibaba/cloud/ai/messagechannel/autoconfigure/MessageChannelAutoConfiguration.java
```diff
@@ -0,0 +1,140 @@
...
+			type = name;
+		}
+		return switch (type.toLowerCase()) {
+			case "dingtalk" -> new DingTalkChannelAdapter(
+					name,
+					requireOption(config, "app-secret", name),
+					config.getOptions().get("webhook-url"),
+					objectMapper,
+					restClient);
+			default -> null;
```
oss-sentinel-ai, line 119: An unsupported configured type is silently omitted, leaving a configuration that only fails later as a 404. Fail startup unless a user adapter with the configured channel name is present, so platform-name typos are actionable.

oss-sentinel-ai (changes requested):
﻿### Summary
Adds `spring-ai-alibaba-starter-message-channel`: a unified IM-channel layer (neutral `ChannelMessage`/`ChannelReply` models, a `MessageChannelAdapter` contract, adapter/agent registries, a dispatcher that threads `channelName:conversationId` as `RunnableConfig.threadId`, `MessagePublisher`, and a single `POST /channel/{name}/callback` webhook), plus a reference DingTalk adapter, Spring Boot auto-configuration, 18 unit tests, and a runnable example. The core security design is sound: inbound HMAC-SHA256 verification runs before payload parsing with constant-time comparison and a ±5-minute replay window, the registries are concurrency-safe (`Map.copyOf` snapshots), user adapters override built-ins by name, and the tests cover the main signature-failure paths. However, several issues should be addressed before publishing this as a production starter — see inline comments.

### Suggestions
1. In the example, make `DINGTALK_APP_SECRET` default to empty/unresolved so startup fails when it is unset, instead of falling back to a publicly known key; consider gating the example behind an explicit profile.
2. Add a dedicated `webhook-secret` property for outbound custom-robot signing, and reject targeted `push(userId, conversationId, ...)` calls (or implement OAuth targeting) rather than silently broadcasting to the group webhook.
3. Move agent-binding resolution inside the dispatcher's try block, and return a generic user-safe failure reply to the IM user while keeping exception details in logs only.
4. Surface async push failures as retryable (non-2xx) or hand delivery to a durable queue, instead of acknowledging with HTTP 200.
5. Consider `@ConditionalOnMissingBean` for the infrastructure beans and gating MVC wiring with `@ConditionalOnWebApplication(type = SERVLET)` so the starter behaves in non-web contexts.

---

PR #4645, opened 2026-05-23
fix(graph): preserve context for compiled subgraphs

### LINKED ISSUE #4191: [BUG] RunnableConfig context data can't passes between nodes

### Current Behavior

Image

Image

### Expected Behavior

Comparing to metadata, context is mutable during execution. It passes information between nodes. Different from OverAllState, context is specific to a run, it will not be persisted.

### Steps To Reproduce

subGraphRunnableConfig.clearContext() -- maybe it should be removed?
Image

### Environment

```markdown
Spring AI Alibaba version(s):
1.1.2.0
```

### Debug logs

_No response_

### Anything else?

_No response_

yu-wubbalubbadubdub:
## Summary
- preserve parent `RunnableConfig.context()` when executing compiled subgraphs
- keep subgraph checkpoint/next-node isolation while allowing run-scoped context to be read inside the subgraph
- add a regression test for parent context flowing into a compiled subgraph node

Fixes #4191

## Validation
- `./mvnw.cmd -pl spring-ai-alibaba-graph-core "-Dtest=CompiledSubGraphTest" test`

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/internal/node/SubCompiledGraphNodeAction.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/internal/node/SubCompiledGraphNodeAction.java
@@ -69,7 +69,6 @@ public record SubCompiledGraphNodeAction(String nodeId, CompileConfig parentComp
 		}).orElse(false);
 
 		RunnableConfig subGraphRunnableConfig = RunnableConfig.builder(config).checkPointId(null).nextNode(null).build();
-		subGraphRunnableConfig.clearContext();
 
 		var parentSaver = parentCompileConfig.checkpointSaver();
 		var subGraphSaver = subGraph.compileConfig.checkpointSaver();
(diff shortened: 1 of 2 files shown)


spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/internal/node/SubCompiledGraphNodeAction.java
```diff
@@ -69,7 +69,6 @@ public CompletableFuture<Map<String, Object>> apply(OverAllState state, Runnable
 		}).orElse(false);
 
 		RunnableConfig subGraphRunnableConfig = RunnableConfig.builder(config).checkPointId(null).nextNode(null).build();
-		subGraphRunnableConfig.clearContext();
```
yuluo-yx, line 72: 这里 clear 会有别的问题，需要兼容考量下

yu-wubbalubbadubdub:
Addressed the compatibility concern from @yuluo-yx.

The subgraph still receives parent run-scoped context, but it receives an isolated copy through RunnableConfig.builder(config). This keeps parent-to-child context propagation while preventing writes inside the subgraph from leaking back into later parent nodes.

Added regression coverage for:
1. parent context value request-id is visible inside the compiled subgraph;
2. child-only context written inside the subgraph is not visible to the following parent node.

Local validation passed:
./mvnw.cmd -pl spring-ai-alibaba-graph-core -Dtest=CompiledSubGraphTest test

PR #4675, opened 2026-06-01
fix(graph): prevent filesystem store path traversal

### LINKED ISSUE #4674: [BUG] FileSystemStore path traversal via namespace/key allows access outside rootDirectory

### Current Behavior

FileSystemStore builds item file paths directly from namespace and key:

private Path createItemPath(List<String> namespace, String key) {
    Path path = rootPath;
    for (String ns : namespace) {
        path = path.resolve(ns);
    }
    return path.resolve(key + ".json");
}

The current validation appears to only check whether namespace and key are null or empty. It does not reject path traversal segments such as .., absolute paths, or path separators.

As a result, crafted namespace or key values can resolve outside the configured rootDirectory.

For example, with namespace = List.of("..") and key = "escaped", the item path resolves to:

root/../escaped.json

which is outside the configured rootDirectory.

This affects putItem, getItem, and deleteItem, because they rely on the same item path construction logic.

I am reporting this as a reproducible file-system storage boundary bug. I am not claiming direct RCE.

### Expected Behavior

FileSystemStore should reject unsafe namespace or key values, or otherwise ensure that the resolved item path always stays under the configured rootDirectory.

putItem, getItem, and deleteItem should not be able to read, write, or delete files outside rootDirectory.

### Steps To Reproduce

Path parent = Files.createTempDirectory("saa-store-parent-");
Path root = parent.resolve("root");

FileSystemStore store = new FileSystemStore(root);

Path escaped = parent.resolve("escaped.json");
System.out.println(Files.exists(escaped)); // false

store.putItem(StoreItem.of(
    List.of(".."),
    "escaped",
    Map.of("x", "y")
));

System.out.println(Files.exists(escaped)); // true

Actual result:

escaped.json is created under the parent directory, outside the configured rootDirectory.

Expected result:

The operation should be rejected, or the resolved item path should be constrained to stay under rootDirectory.

### Environment

```markdown
Module: spring-ai-alibaba-graph-core
Class: com.alibaba.cloud.ai.graph.store.stores.FileSystemStore
```

### Debug logs

_No response_

### Anything else?

I am willing to submit a PR to fix this issue.

SPUERSAIYAN:
## Describe what this PR does / why we need it

This PR fixes a path traversal issue in `FileSystemStore`.

Previously, `FileSystemStore` built item file paths directly from `namespace` and `key`. Crafted values such as `namespace = List.of("..")` or traversal segments in `key` could cause the resolved item path to escape the configured `rootDirectory`.

This PR ensures that item paths are validated before file operations so that `putItem`, `getItem`, and `deleteItem` cannot access files outside the configured store root.

## Does this pull request fix one issue?

Fixes #4674

## Describe how you did it

- Normalize the store root path.
- Normalize the resolved item path before using it.
- Validate that the final item path stays under the configured `rootDirectory`.
- Reject unsafe paths that would escape the store root.
- Add regression tests for path traversal through `namespace` and `key`.

## Describe how to verify it

Run the related tests:

```bash
./mvnw -pl :spring-ai-alibaba-graph-core -Dtest=FileSystemStoreTest,StoreIntegrationTest,GraphStoreIntegrationTest test

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStore.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStore.java
@@ -57,7 +57,7 @@ public class FileSystemStore extends BaseStore {
 	 * @param rootPath the root path for storage
 	 */
 	public FileSystemStore(Path rootPath) {
-		this.rootPath = rootPath;
+		this.rootPath = rootPath.toAbsolutePath().normalize();
 		this.objectMapper = new ObjectMapper();
 		this.objectMapper.findAndRegisterModules();
 		initializeRootDirectory();
@@ -74,6 +74,9 @@ public class FileSystemStore extends BaseStore {
 
 			String itemJson = objectMapper.writeValueAsString(item);
 			Files.write(itemPath, itemJson.getBytes());
+		}
+		catch (IllegalArgumentException e) {
+			throw e;
 		}
 		catch (Exception e) {
 			throw new RuntimeException("Failed to store item to file system", e);
@@ -98,6 +101,9 @@ public class FileSystemStore extends BaseStore {
 			StoreItem item = objectMapper.readValue(itemJson, StoreItem.class);
 			return Optional.of(item);
 		}
+		catch (IllegalArgumentException e) {
+			throw e;
+		}
 		catch (Exception e) {
 			throw new RuntimeException("Failed to retrieve item from file system", e);
 		}
@@ -120,6 +126,9 @@ public class FileSystemStore extends BaseStore {
 				return true;
 			}
 			return false;
+		}
+		catch (IllegalArgumentException e) {
+			throw e;
 		}
 		catch (Exception e) {
 			throw new RuntimeException("Failed to delete item from file system", e);
@@ -245,7 +254,11 @@ public class FileSystemStore extends BaseStore {
 		for (String ns : namespace) {
 			path = path.resolve(ns);
 		}
-		return path.resolve(key + ".json");
+		Path itemPath = path.resolve(key + ".json").toAbsolutePath().normalize();
+		if (!itemPath.startsWith(rootPath)) {
+			throw new IllegalArgumentException("store item path must stay under root directory");
+		}
+		return itemPath;
 	}
 
 	/**
--- a/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStoreTest.java
+++ b/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStoreTest.java
@@ -145,6 +145,34 @@ class FileSystemStoreTest {
 	}
 
 	@Test
+	void testRejectsPathTraversalInNamespace() {
+		Path escapedFile = tempDir.resolveSibling("escaped.json");
+
+		assertThrows(IllegalArgumentException.class,
+				() -> store.putItem(StoreItem.of(List.of(".."), "escaped", Map.of("data", "test_value"))));
+
+		assertThat(Files.exists(escapedFile)).isFalse();
+	}
+
+	@Test
+	void testRejectsPathTraversalInKey() {
+		Path escapedFile = tempDir.resolveSibling("escaped.json");
+
+		assertThrows(IllegalArgumentException.class,
+				() -> store.putItem(StoreItem.of(List.of("safe"), "../../escaped", Map.of("data", "test_value"))));
+
+		assertThat(Files.exists(escapedFile)).isFalse();
+	}
+
+	@Test
+	void testRejectsPathTraversalForReadAndDelete() {
+		assertThrows(IllegalArgumentException.class, () -> store.getItem(List.of(".."), "escaped"));
+		assertThrows(IllegalArgumentException.class, () -> store.getItem(List.of("safe"), "../../escaped"));
+		assertThrows(IllegalArgumentException.class, () -> store.deleteItem(List.of(".."), "escaped"));
+		assertThrows(IllegalArgumentException.class, () -> store.deleteItem(List.of("safe"), "../../escaped"));
+	}
+
+	@Test
 	void testSizeAndClear() {
 		// Given
 		assertThat(store.isEmpty()).isTrue();

### CODE REVIEW

**Comment by SPUERSAIYAN** [comment]
Hi maintainers, this is my first PR to this project. I’m very glad to have the opportunity to contribute to Spring AI Alibaba. The workflows are currently waiting for maintainer approval. Thanks a lot for your time and review!

File: spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStore.java [hunk 1]
```diff
@@ -246,24 +255,13 @@ private Path createItemPath(List<String> namespace, String key) {
 			validatePathSegment(ns, "namespace");
 			path = path.resolve(ns);
 		}
-		validatePathSegment(key, "key");
-		Path itemPath = path.resolve(key + ".json").normalize();
-		if (!itemPath.startsWith(rootPath.toAbsolutePath().normalize())) {
-			throw new IllegalArgumentException("resolved path escapes root directory");
+		Path itemPath = path.resolve(key + ".json").toAbsolutePath().normalize();
```
oss-sentinel-ai, line 258: Removing `validatePathSegment(key, ...)` changes the security semantics: a single-level traversal key such as `"../escaped"` now resolves *inside* the root (crossing namespace boundaries — `putItem(List.of("safe"), "../escaped")` would write to `<root>/escaped.json`) instead of being rejected, while multi-level traversal (`"../../x"`) is still rejected by the containment check below. This breaks the pre-existing test `FileSystemStoreTest.testRejectsKeyTraversalSegments`, which expects an `IllegalArgumentException` containing "unsafe path segment", and is very likely why CI is currently failing on this PR. Suggestion: keep per-key segment validation (reject `..`, path separators, absolute paths, empty) and treat the containment check as defense-in-depth.

spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStore.java
```diff
@@ -75,6 +75,9 @@ public void putItem(StoreItem item) {
 			String itemJson = objectMapper.writeValueAsString(item);
 			Files.write(itemPath, itemJson.getBytes());
 		}
+		catch (IllegalArgumentException e) {
```
oss-sentinel-ai, line 78: Rethrowing `IllegalArgumentException` unwrapped changes the exception contract: the pre-existing test `FileSystemStoreTest.testRejectsNamespaceTraversalSegments` asserts a `RuntimeException` whose *cause* is an `IllegalArgumentException` (the old code wrapped it via `new RuntimeException(msg, e)`). With this change the raw `IllegalArgumentException` propagates with no cause, so that assertion fails. Either keep the wrapping or update the two pre-existing tests in this PR to the new contract. The same applies to the added catch blocks in `getItem`/`deleteItem`.

spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStore.java
```diff
@@ -246,24 +255,13 @@ private Path createItemPath(List<String> namespace, String key) {
 			validatePathSegment(ns, "namespace");
 			path = path.resolve(ns);
 		}
-		validatePathSegment(key, "key");
-		Path itemPath = path.resolve(key + ".json").normalize();
-		if (!itemPath.startsWith(rootPath.toAbsolutePath().normalize())) {
-			throw new IllegalArgumentException("resolved path escapes root directory");
+		Path itemPath = path.resolve(key + ".json").toAbsolutePath().normalize();
+		if (!itemPath.startsWith(rootPath)) {
+			throw new IllegalArgumentException("store item path must stay under root directory");
```
oss-sentinel-ai, line 260: The containment check alone accepts keys containing `/` (e.g. `"a/b"` creates nested directories under the namespace) and empty keys (producing a file named `.json`); both were rejected by the previous validation. If this relaxation is intentional, please pin it down with tests; otherwise reject path separators in keys explicitly.

spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/store/stores/FileSystemStoreTest.java
```diff
@@ -144,6 +144,34 @@ void testValidationErrors() {
...
+		Path escapedFile = tempDir.resolveSibling("escaped.json");
+
+		assertThrows(IllegalArgumentException.class,
+				() -> store.putItem(StoreItem.of(List.of("safe"), "../../escaped", Map.of("data", "test_value"))));
+
+		assertThat(Files.exists(escapedFile)).isFalse();
+	}
+
+	@Test
+	void testRejectsPathTraversalForReadAndDelete() {
```
oss-sentinel-ai, line 168: Nice traversal coverage across put/get/delete. Suggested additions: an absolute key (e.g. `"/etc/passwd"`), an empty key, and a key with an embedded slash — their behavior is now decided solely by the containment logic and is worth pinning down.

oss-sentinel-ai (changes requested):
### Summary
This PR replaces per-segment key validation in `FileSystemStore` with constructor-side root normalization plus a final containment check, and adds traversal tests. The containment approach is sound as defense-in-depth, but removing key-segment validation and unwrapping the rethrown `IllegalArgumentException` changes the security semantics and the exception contract, breaking two pre-existing tests (`testRejectsKeyTraversalSegments`, `testRejectsNamespaceTraversalSegments`) — which is very likely why CI is currently unstable on this PR.

### Findings
- **[Critical]** FileSystemStore.java:258 — single-level key traversal (`"../x"`) is now allowed inside root and crosses namespace boundaries; breaks existing `testRejectsKeyTraversalSegments`
- **[Critical]** FileSystemStore.java:78 — unwrapped `IllegalArgumentException` rethrow breaks `testRejectsNamespaceTraversalSegments` (expects RuntimeException with IAE cause)
- **[Warning]** FileSystemStore.java:260 — containment-only check accepts keys with `/` and empty keys, previously rejected
- **[Info]** FileSystemStoreTest.java:168 — suggest adding absolute-key / empty-key / embedded-slash cases

### Suggestions
Keep `validatePathSegment(key, "key")` (reject `..`, path separators, absolute paths, empty) in addition to the containment check, and either preserve the wrapped-exception contract or update the two pre-existing tests in this PR. Then re-run `./mvnw -pl spring-ai-alibaba-graph-core test -Dtest=FileSystemStoreTest` to confirm the whole class passes.

---

PR #4688, opened 2026-06-03
fix: correct ConversationChatMemory eviction logic

### LINKED ISSUE #4687: [BUG] ConversationChatMemory evicts messages based on incoming batch size instead of stored history size

### Current Behavior

`ConversationChatMemory` is intended to keep conversation history within the configured cache limit from `CommonConfig#getMaxConversationRoundInCache()`.

However, `ConversationChatMemory#add` currently decides whether to evict old messages by checking the incoming batch size:

```java
for (Message message : messages) {
    if (messages.size() >= maxMessages) {
        historyMessages.poll();
    }

    historyMessages.offer(message);
}

```
`messages.size()` only represents the number of messages in the current add call, not the accumulated history size loaded from Redis.

As a result, when messages are appended in small batches, especially one message at a time, the cached conversation history can grow beyond `maxMessages` and old messages are not evicted.

This can happen in normal agent memory usage. For example, Spring AI's `MessageChatMemoryAdvisor` adds the user message and assistant message through separate `ChatMemory#add` calls, so the incoming batch size is often 1.

### Expected Behavior

Eviction should be based on the accumulated stored history size, not the incoming batch size.

For example, if the configured limit is 3, after adding 5 messages sequentially to the same conversationId, only the latest 3 messages should remain in Redis.

### Steps To Reproduce

1. Configure `CommonConfig#getMaxConversationRoundInCache()` to return 3.
2. Create a `ConversationChatMemory` instance.
3. Add 5 messages to the same `conversationId`, one message per add() call.
4. Call `get(conversationId)`.
Expected result: 3 messages are returned.

Actual result: 5 messages are returned because messages.size() is 1 for each call, so the eviction condition is never triggered.

### Environment

```markdown
Spring AI Alibaba version(s): main branch
Commit: 8c78cf6dc506731fc2f8647a52ad31ae863ff463
Module: spring-ai-alibaba-admin-server-core
Java: 17+
```

### Debug logs

_No response_

### Anything else?

The eviction logic probably should check `historyMessages.size()` after adding messages and remove old messages while the stored history exceeds the configured limit.

One detail worth confirming is whether `maxConversationRoundInCache` is expected to represent conversation rounds or raw messages, since the implementation currently stores individual `Message` objects and names the internal field `maxMessages`. In either case, the current condition should not be based on `messages.size()`.

yin-bo-Final:
This PR fixes `ConversationChatMemory` eviction logic.

Previously, `ConversationChatMemory#add` checked the incoming batch size with `messages.size()` to decide whether old messages should be evicted. However, `messages.size()` only represents the number of messages in the current `add()` call, not the accumulated conversation history loaded from Redis.

As a result, when messages are appended one by one, cached conversation history can grow beyond the configured limit and old messages are not evicted as expected.

Fixes #4687

### Describe how you did it
Changed the eviction logic to append each new message first, then trim old messages while the stored history size exceeds `maxMessages`.

Also added unit tests for:

- sequential/small-batch append where stored history already exceeds the limit
- large incoming batch append where only the latest messages should be kept

### Describe how to verify it
Run:
```bash
cd spring-ai-alibaba-admin
../mvnw -pl spring-ai-alibaba-admin-server-core -am -Dtest=ConversationChatMemoryTest test
```
Note: I tried this command locally, but the current `spring-ai-alibaba-admin-server-runtime` module fails to compile before reaching this test due to existing unrelated errors in `ToolExample.java` and runtime module dependencies.
### Special notes for reviews
The fix keeps the existing `maxMessages` behavior and does not change the meaning of `maxConversationRoundInCache`. It only corrects the eviction condition so it is based on stored history size instead of incoming batch size.

### DIFF
--- a/spring-ai-alibaba-admin/spring-ai-alibaba-admin-server-core/src/main/java/com/alibaba/cloud/ai/studio/core/agent/memory/ConversationChatMemory.java
+++ b/spring-ai-alibaba-admin/spring-ai-alibaba-admin-server-core/src/main/java/com/alibaba/cloud/ai/studio/core/agent/memory/ConversationChatMemory.java
@@ -64,12 +64,13 @@ public class ConversationChatMemory implements ChatMemory {
 			historyMessages = new ArrayDeque<>();
 		}
 
+		int messageLimit = Math.max(0, maxMessages);
 		for (Message message : messages) {
 			historyMessages.offer(message);
+		}
 
-			while (historyMessages.size() > maxMessages) {
-				historyMessages.poll();
-			}
+		while (historyMessages.size() > messageLimit) {
+			historyMessages.poll();
 		}
 
 		redisManager.put(key, historyMessages);
--- a/spring-ai-alibaba-admin/spring-ai-alibaba-admin-server-core/src/main/java/com/alibaba/cloud/ai/studio/core/agent/memory/ConversationChatMemory.java
+++ b/spring-ai-alibaba-admin/spring-ai-alibaba-admin-server-core/src/main/java/com/alibaba/cloud/ai/studio/core/agent/memory/ConversationChatMemory.java
@@ -65,11 +65,11 @@ public class ConversationChatMemory implements ChatMemory {
 		}
 
 		for (Message message : messages) {
-			if (messages.size() >= maxMessages) {
+			historyMessages.offer(message);
+
+			while (historyMessages.size() > maxMessages) {
 				historyMessages.poll();
 			}
-
-			historyMessages.offer(message);
 		}
 
 		redisManager.put(key, historyMessages);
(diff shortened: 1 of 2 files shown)


yin-bo-Final:
Thanks for the review. I addressed the feedback in the latest commit.

Changes made:
- Normalized maxMessages to a non-negative value before trimming, so negative configuration values cannot cause an infinite loop.
- Moved the eviction loop to trim based on the final stored history size after appending messages.
- Updated tests to assert the persisted Redis argument via ArgumentCaptor instead of relying on the same Deque instance.
- Added a regression test for negative maxMessages.

Please take another look when you have time.

yuluo-yx (approved):
LGTM thx

PR #4689, opened 2026-06-03
fix(studio): multi-Agent response issues and add Windows chat-ui build script

xs-java2ai:

1.在多Agent特别是并行执行（Parallel Agent）中，studio响应混乱
解决：加入按subAgent进行响应UI
image

2. 缺少windows下的编译脚本，只有linux环境的sh文件
解决：加入windows下的spring-ai-alibaba-studio/agent-chat-ui/build-and-deploy.bat编译脚本


GTyingzi:
@codex

spring-ai-alibaba-studio/src/main/java/com/alibaba/cloud/ai/agent/studio/controller/ExecutionController.java
```diff
@@ -358,4 +358,20 @@ else if (nodeOutput instanceof InterruptionMetadata interruptionMetadata) {
...
 	}
+
+	/**
+	 * Checks if a StreamingOutput comes from a sub-graph (sub-agent) by reflectively
+	 * calling {@code isSubGraph()} on the StreamingOutput instance. Returns false if
+	 * the method is not available or invocation fails.
+	 */
+	private boolean isSubGraphOutput(StreamingOutput<?> streamingOutput) {
+		try {
+			java.lang.reflect.Method method = streamingOutput.getClass().getMethod("isSubGraph");
```
oss-sentinel-ai, line 369: Reflection is unnecessary here: `isSubGraph()` is a public method on `NodeOutput` (the direct superclass of `StreamingOutput`). Replace this entire method with `return streamingOutput.isSubGraph();`. The reflective approach silently swallows all exceptions and returns false, which could mask real errors and makes debugging harder.

spring-ai-alibaba-studio/src/main/java/com/alibaba/cloud/ai/agent/studio/controller/ExecutionController.java
```diff
@@ -280,18 +277,23 @@ private Flux<ServerSentEvent<String>> executeAgent(UserMessage userMessage, Agen
...
 							}
 							else {
-//						chunkBuilder.append(assistantMessage.getText());
 								agentResponse = new AgentRunResponse(node, agentName, assistantMessage, tokenUsage, assistantMessage.getText());
 							}
 						}
 						else {
 							agentResponse = new AgentRunResponse(node, agentName, message, tokenUsage, "");
 						}
+						if (agentResponse != null) {
```
oss-sentinel-ai, line 294: The null check `if (agentResponse != null)` is redundant. At this point in the code, `agentResponse` is always non-null because the `message == null` case returns early at line 276, and all subsequent branches assign `agentResponse`.

spring-ai-alibaba-studio/src/main/java/com/alibaba/cloud/ai/agent/studio/dto/messages/AgentRunResponse.java
```diff
@@ -42,6 +44,24 @@ public class AgentRunResponse {
...
+	@JsonInclude(JsonInclude.Include.NON_DEFAULT)
+	private boolean subGraph;
+
+	/**
+	 * Parallel sub-agent results. Present only when this response comes from a
+	 * ParallelAgent node whose state contains multiple named AssistantMessage results.
+	 * Key = sub-agent result name (e.g. "poem_result"), Value = sub-agent result.
+	 */
+	@JsonProperty("subAgents")
+	@JsonInclude(JsonInclude.Include.NON_NULL)
```
oss-sentinel-ai, line 62: The `subAgents` field and the `SubAgentResult` inner class (lines 169-212) are never populated by `ExecutionController` or any other code. The parallel grouping is done entirely client-side in `Stream.tsx`. This is dead code that adds API surface and a public constructor (line 93) without any producer. Either remove it or wire it up in the controller.

spring-ai-alibaba-studio/agent-chat-ui/src/providers/Stream.tsx
```diff
@@ -202,7 +205,56 @@ export const StreamProvider: React.FC<StreamProviderProps> = ({ children }) => {
...
+              subAgents,
+              metadata: {},
+            };
+
+            setMessages(prev => {
+              const newMessages = [...prev];
+              // Find existing parallel message (last one)
+              let parallelIdx = -1;
+              for (let i = newMessages.length - 1; i >= 0; i--) {
+                if (newMessages[i].message.messageType === 'parallel') {
```
oss-sentinel-ai, line 237: Searching backward through ALL messages for the last `parallel` type will corrupt earlier parallel messages in multi-turn conversations. If a user sends two messages that each trigger parallel agents, the second run finds and overwrites the first parallel message's accumulated content. The fix: track the parallel message ID locally (like `isFirstChunk`) and only update that specific message, or scope the search to messages added after the current user message.

spring-ai-alibaba-studio/agent-chat-ui/src/providers/Stream.tsx
```diff
@@ -202,7 +205,56 @@ export const StreamProvider: React.FC<StreamProviderProps> = ({ children }) => {
             continue;
           }
 
-          // Use chunk for streaming updates if available
+          // ── Sub-agent (subGraph) streaming chunk handling ──
+          if (agentResponse.chunk && (agentResponse as AgentRunResponse).subGraph) {
```
oss-sentinel-ai, line 209: The `resumeFeedback` callback (lines 432-535) was not updated to handle `subGraph` chunks. If a tool-resume triggers parallel agents, chunks will fall through to the normal path and be interleaved into a single message — the same bug this PR fixes for `sendMessage`. The subGraph handling block should be duplicated or extracted into a shared helper.

spring-ai-alibaba-studio/agent-chat-ui/src/providers/Stream.tsx
```diff
@@ -202,7 +205,56 @@ export const StreamProvider: React.FC<StreamProviderProps> = ({ children }) => {
...
+              // Find existing parallel message (last one)
+              let parallelIdx = -1;
+              for (let i = newMessages.length - 1; i >= 0; i--) {
+                if (newMessages[i].message.messageType === 'parallel') {
+                  parallelIdx = i;
+                  break;
+                }
+              }
+              const uiMsg: UIMessage = {
+                id: `parallel-${Date.now()}`,
```
oss-sentinel-ai, line 243: Using `Date.now()` for ID generation can produce duplicate IDs if React re-renders multiple chunks within the same millisecond during rapid streaming. Consider using a monotonically increasing counter or combining `Date.now()` with the sub-agent name.

spring-ai-alibaba-studio/agent-chat-ui/src/components/thread/messages/parallel.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+      <div
+        className={cn(
+          "grid gap-3",
+          subAgents.length === 2 && "grid-cols-1 sm:grid-cols-2",
+          subAgents.length >= 3 && "grid-cols-1 sm:grid-cols-2 lg:grid-cols-3"
+        )}
+      >
+        {subAgents.map((sub, idx) => (
+          <div
```
oss-sentinel-ai, line 53: `key={sub.name || idx}` can produce duplicate React keys if two sub-agents share the same name. Consider always using `idx` or combining name with index: `key={\`${sub.name}-${idx}\`}`.

spring-ai-alibaba-studio/agent-chat-ui/build-and-deploy.bat
```diff
@@ -0,0 +1,47 @@
...
+
+set STATIC_EXPORT=true
+echo   Running npx next build ...
+call npx next build
+if %ERRORLEVEL% neq 0 (
+    echo ERROR: Build failed
+    exit /b 1
+)
+
+if not exist "%OUT_DIR%" (
```
oss-sentinel-ai, line 35: The shell script uses `pnpm run build:static` but this bat file uses `npx next build` with `STATIC_EXPORT=true`. These may produce different builds depending on what `build:static` configures in `package.json` vs a raw `next build`. This should match the shell script's behavior for consistency.

spring-ai-alibaba-studio/agent-chat-ui/build-and-deploy.bat
```diff
@@ -0,0 +1,47 @@
...
+    echo ERROR: Build output directory not found
+    exit /b 1
+)
+
+echo   Build complete
+
+echo Step 3: Copying files to target directory...
+xcopy /e /i /q /y "%OUT_DIR%\*" "%TARGET_DIR%"
+
```
oss-sentinel-ai, line 44: No error check after `xcopy`. If the copy fails, the script still prints 'Deploy complete' and exits with success. Add `if %ERRORLEVEL% neq 0` after the xcopy command, matching the pattern used after the build step.

spring-ai-alibaba-studio/agent-chat-ui/build-and-deploy.bat
```diff
@@ -0,0 +1,47 @@
...
+echo Step 1: Cleaning target directory...
+if exist "%TARGET_DIR%" (
+    rd /s /q "%TARGET_DIR%"
+    mkdir "%TARGET_DIR%"
+    echo   Target directory cleaned
+) else (
+    mkdir "%TARGET_DIR%"
+    echo   Created target directory
+)
```
oss-sentinel-ai, line 20: The bat script does `rd /s /q` (full directory delete then recreate) while the shell script does `rm -rf "$TARGET_DIR"/*` (contents only). If `TARGET_DIR` is misconfigured, `rd /s /q` could delete a parent directory. Consider adding a sanity check that `TARGET_DIR` contains the expected path segments before deleting.

spring-ai-alibaba-studio/agent-chat-ui/src/types/messages.ts
```diff
@@ -67,13 +67,29 @@ export interface ToolResponse {
   responseData: string;
 }
 
+// Parallel sub-agent result
+export interface SubAgentResult {
```
oss-sentinel-ai, line 71: The frontend `SubAgentResult` type has a `content: string` field, but the Java `SubAgentResult` inner class has `chunk: String` (not `content`). While the backend never populates this field today, the schema mismatch means they will not interoperate if the backend starts sending `subAgents` in the future.

oss-sentinel-ai (changes requested):
### Summary
Fixes garbled multi-agent (ParallelAgent) streaming output by adding a `subGraph` flag to backend SSE responses and grouping chunks per sub-agent client-side, plus a Windows variant of the chat-ui build script. The approach is sound and backward-compatible, but a state-tracking bug corrupts parallel messages across conversation turns, and the resume path is left unfixed — requesting changes before merge.

### Findings
- **[Critical]** `agent-chat-ui/src/providers/Stream.tsx:237` — backward search for the last `parallel` message scans ALL messages, so a second parallel run in a session overwrites the first parallel message's accumulated content.
- **[Warning]** `agent-chat-ui/src/providers/Stream.tsx:209` — `resumeFeedback` does not handle `subGraph` chunks; resume-triggered parallel runs are still interleaved (the same bug this PR fixes for `sendMessage`).
- **[Warning]** `ExecutionController.java:369` — unnecessary reflection for `isSubGraph()`; the method is public and directly callable, and the reflection silently swallows exceptions.
- **[Warning]** `AgentRunResponse.java:62` — `subAgents`/`SubAgentResult` is dead API surface never populated by any producer; also mismatches the frontend type (`chunk` vs `content`).
- **[Warning]** `build-and-deploy.bat` — diverges from `build-and-deploy.sh`: `npx next build` vs `pnpm run build:static`, no ERRORLEVEL check after `xcopy`, and `rd /s /q` on the target dir without a sanity check.
- **[Info]** `Stream.tsx:243` / `parallel.tsx:53` — `Date.now()` IDs can collide within the same millisecond; `key={sub.name || idx}` can produce duplicate React keys for same-named sub-agents.
- **[Info]** No tests cover the changed controller logic; the `subGraph` flag and `isSubGraphOutput` path have zero coverage.

### Suggestions
- Track the in-flight parallel message ID locally (like `isFirstChunk`) or scope grouping to messages created after the current user message, instead of a global backward search.
- Extract the subGraph chunk handling into a shared helper used by both `sendMessage` and `resumeFeedback`.
- Replace the reflection with `streamingOutput.isSubGraph()`; either remove the unused `subAgents` field or populate it in the controller.
- Align the `.bat` with the shell script: same build command, ERRORLEVEL check after every step, and guard the directory deletion.

---

PR #4736, opened 2026-06-20
fix(admin): guard null topK/threshold in KnowledgeBaseDocumentRetriever

### LINKED ISSUE #4734: [BUG] [admin] Agent chat with a knowledge base throws NPE: FileSearchOptions.getTopK() is null (#3867 regression)

### Current Behavior

  Chatting with an Agent app that has a knowledge base attached throws an NPE on the first message (during RAG retrieval), so RAG-backed
  chat is unusable whenever the agent-level FileSearchOptions.topK is not explicitly set:

  java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of
  "...FileSearchOptions.getTopK()" is null
      at ...core.rag.retriever.KnowledgeBaseDocumentRetriever.retrieve(KnowledgeBaseDocumentRetriever.java:102)
      at ...core.rag.advisor.KnowledgeBaseRetrievalAdvisor.before(KnowledgeBaseRetrievalAdvisor.java:128)
  (logged as "Aggregation Error" / "Stream processing failed" wrapping the NPE)

### Expected Behavior

  Agent chat with a knowledge base should retrieve docs and reply normally; an unset topK should fall back to a default (FileSearchOptions
  Javadoc says "default to 3"), not NPE.

### Steps To Reproduce

  1. Start admin backend + frontend.
  2. Create an Agent app, attach a knowledge base (RAG enabled).
  3. Leave the agent-level file-search topK unset (default).
  4. Open the app chat, send any message.
  5. NPE in backend log; chat fails.

### Environment

```markdown
Spring AI Alibaba version(s):
  Spring AI Alibaba version(s): current main; module spring-ai-alibaba-admin-server-core (RAG retriever); spring-ai 1.1.2; introduced by #3867.
```

### Debug logs

  Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of
  "com.alibaba.cloud.ai.studio.runtime.domain.app.FileSearchOptions.getTopK()" is null
      at com.alibaba.cloud.ai.studio.core.rag.retriever.KnowledgeBaseDocumentRetriever.retrieve(KnowledgeBaseDocumentRetriever.java:102)
      at com.alibaba.cloud.ai.studio.core.rag.advisor.KnowledgeBaseRetrievalAdvisor.before(KnowledgeBaseRetrievalAdvisor.java:128)
  (wrapped in IllegalStateException: Stream processing failed ... MessageAggregator Aggregation Error)

### Anything else?

  Root cause: the public KnowledgeBaseDocumentRetriever.retrieve(Query) auto-unboxes getTopK()/getSimilarityThreshold() without null
  checks, while the private retrieve(KnowledgeBase, Query) in the SAME file already null-checks both — an inconsistency.
  FileSearchOptions.topK/similarityThreshold have no @Builder.Default despite their Javadoc claiming defaults (3 / 0.2), so they're null
  when unset.

  Suggested fix (null-safe, mirrors the private method — null topK = no post-merge cap, null threshold = skip filter):
      Integer topK = searchOptions.getTopK();
      Float similarityThreshold = searchOptions.getSimilarityThreshold();
      ... .filter(x -> x.getScore() != null && (similarityThreshold == null || x.getScore() > similarityThreshold))
          .limit(topK != null ? topK.longValue() : Long.MAX_VALUE)

  Companion hardening: give FileSearchOptions.topK/similarityThreshold @Builder.Default (3 / 0.2f) so unset options don't silently become
  null.

  This is the 3rd bug I've hit that traces to #3867 (also: legacy Playground currentPrompt ReferenceError, Date.now() React key collision)
  — worth a regression pass over the RAG + Playground paths that PR added.

  I have a ready patch + can open a PR.

XianReallyHot-ZZH:

Fixes the NPE reported in #4734: chatting with an Agent app that has a knowledge base throws `NullPointerException: ...FileSearchOptions.getTopK() is null` on the first turn whenever the agent-level `topK` / `similarityThreshold` are not explicitly set (the default), making RAG-backed chat unusable.

Fixes #4734

### Describe how you did it

In `KnowledgeBaseDocumentRetriever`, the public `retrieve(Query)` auto-unboxes `searchOptions.getSimilarityThreshold()` (in the `.filter`) and `searchOptions.getTopK()` (in the `.limit`), so either being `null` throws. The private `retrieve(KnowledgeBase, Query)` in the same class already null-checks both; this PR mirrors that null-safety in the public method:

- `null` `topK` → no post-merge cap (`Long.MAX_VALUE`);
- `null` `similarityThreshold` → skip the threshold filter (per-KB retrieval already applied its own).

### Describe how to verify it

1. Build & start the admin backend + frontend.
2. Create an Agent app, attach a knowledge base (RAG enabled), leave the file-search `topK` unset (default).
3. Open the app's chat, send any message → no NPE; chat returns results.

### Special notes for reviews

- Single-file, ~5-line change; mirrors the null-safety already present in the private `retrieve(KnowledgeBase, Query)` of the same class.
- Optional companion (NOT included — maintainer call): `FileSearchOptions.topK` / `similarityThreshold` document defaults of `3` / `0.2` but have no `@Builder.Default`, so they're `null` when unset. Adding `@Builder.Default` would enforce those defaults everywhere; left out here to keep this PR focused on the NPE and because it changes the meaning of "unset" (null = uncapped vs. fixed default). Happy to follow up if you prefer that direction.

### DIFF
--- a/spring-ai-alibaba-admin/spring-ai-alibaba-admin-server-core/src/main/java/com/alibaba/cloud/ai/studio/core/rag/retriever/KnowledgeBaseDocumentRetriever.java
+++ b/spring-ai-alibaba-admin/spring-ai-alibaba-admin-server-core/src/main/java/com/alibaba/cloud/ai/studio/core/rag/retriever/KnowledgeBaseDocumentRetriever.java
@@ -96,10 +96,13 @@ public class KnowledgeBaseDocumentRetriever implements DocumentRetriever {
 				documents.addAll(future.get(SEARCH_TIMEOUT, TimeUnit.SECONDS));
 			}
 
+			Integer topK = searchOptions.getTopK();
+			Float similarityThreshold = searchOptions.getSimilarityThreshold();
 			List<Document> results = documents.stream()
 				.sorted(Comparator.comparing(Document::getScore, Comparator.nullsLast(Comparator.reverseOrder())))
-				.filter(x -> x.getScore() != null && x.getScore() > searchOptions.getSimilarityThreshold())
-				.limit(searchOptions.getTopK())
+				.filter(x -> x.getScore() != null
+						&& (similarityThreshold == null || x.getScore() > similarityThreshold))
+				.limit(topK != null ? topK.longValue() : Long.MAX_VALUE)
 				.toList();
 
 			LogUtils.monitor("DocumentRetriever", "retrieve", start, SUCCESS, query.text(), results.size());


GTyingzi (commented):
LGTM

PR #4798, opened 2026-07-09
fix: honor RetryConfig enable in HttpNode

### LINKED ISSUE #4797: [BUG] HttpNode ignores RetryConfig.enable and still retries disabled requests

### Current Behavior

`HttpNode` ignores `RetryConfig.enable` and still applies retry behavior even when retry is explicitly disabled.

In `HttpNode.apply(...)`, the request pipeline always attaches `retryWhen(...)`:

```java
Mono<ResponseEntity<byte[]>> responseMono = requestSpec
    .exchangeToMono((ClientResponse resp) -> resp.toEntity(byte[].class))
    .retryWhen(Retry.backoff(retryConfig.maxRetries, Duration.ofMillis(retryConfig.maxRetryInterval)));
```
However, RetryConfig has an enable field and isEnable() method, but this flag is not checked during request execution.

As a result, a configuration such as:
```java
new RetryConfig(3, 100, false)
```
still retries failed HTTP requests.

This is especially risky for non-idempotent methods such as POST, PUT, where an unexpected retry may trigger duplicate external side effects.

### Expected Behavior

When `RetryConfig.enable` is `false`, `HttpNode` should not retry the request.

`Retry` behavior should only be `applied` when `retryConfig.isEnable()` returns `true`.

### Steps To Reproduce

1. Create an HttpNode with a RetryConfig whose enable value is false.
2. Use an HTTP endpoint or MockWebServer that fails the first request, for example by disconnecting at the start.
3. Configure the node like this:
```java
HttpNode node = HttpNode.builder()
    .webClient(webClient)
    .method(HttpMethod.POST)
    .url(url)
    .retryConfig(new RetryConfig(3, 100, false))
    .build();
```
4. Execute node.apply(new OverAllState()).
5. Observe that the request is still retried even though retry was disabled.

### Environment

```markdown
Spring AI Alibaba version: latest main branch
```

### Debug logs

No application debug logs are required to reproduce this issue.

A minimal MockWebServer-based reproduction shows the behavior:

- Configure `RetryConfig(3, 100, false)`
- Make the first request fail with `SocketPolicy.DISCONNECT_AT_START`
- Execute `HttpNode.apply(...)`
- The request is still retried even though retry is disabled

Expected request count: `1`  
Actual request count: greater than `1`

### Anything else?

The fix should likely apply retryWhen(...) conditionally, for example only when retryConfig.isEnable() is true.

A regression test can verify that when enable=false, a failed request is attempted only once.

yin-bo-Final:

This PR fixes `HttpNode` retry behavior.

Before this change, `HttpNode` would still retry failed HTTP requests when `RetryConfig.enable` was set to `false`. This could be unexpected, especially for POST/PUT/PATCH requests.

Fixes #4797 

### Describe how you did it

I changed `HttpNode` to only apply `retryWhen(....)`  when  `retryConfig.isEnable()` is true'.

I also added a test to make sure retry-disabled requests are only sent once.

### Describe how to verify it
Run it:

```bash
./mvnw.cmd -pl spring-boot-starters/spring-ai-alibaba-starter-builtin-nodes -Dtest=HttpNodeTest test
```

### Special notes for reviews
None.

### DIFF
--- a/spring-boot-starters/spring-ai-alibaba-starter-builtin-nodes/src/main/java/com/alibaba/cloud/ai/graph/node/HttpNode.java
+++ b/spring-boot-starters/spring-ai-alibaba-starter-builtin-nodes/src/main/java/com/alibaba/cloud/ai/graph/node/HttpNode.java
@@ -143,8 +143,12 @@ public class HttpNode implements NodeAction {
 			initBody(body, requestSpec, state);
 
 			Mono<ResponseEntity<byte[]>> responseMono = requestSpec
-				.exchangeToMono((ClientResponse resp) -> resp.toEntity(byte[].class))
-				.retryWhen(Retry.backoff(retryConfig.maxRetries, Duration.ofMillis(retryConfig.maxRetryInterval)));
+				.exchangeToMono((ClientResponse resp) -> resp.toEntity(byte[].class));
+			if (retryConfig.isEnable()) {
+				responseMono = responseMono.retryWhen(
+						Retry.backoff(retryConfig.maxRetries,Duration.ofMillis(retryConfig.maxRetryInterval)));
+			}
+
 			ResponseEntity<byte[]> responseEntity = responseMono.block();
 			Map<String, Object> httpResponse = processResponse(responseEntity, state);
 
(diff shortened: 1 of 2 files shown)


guanxuc (approved):
LGTM, thanks for your contribution!

PR #4803, opened 2026-07-10
fix(agent): parse YAML list tools in agent specs

chensishang:
## Summary
  - Fix `AgentSpecLoader` so `tools` in YAML front matter can be parsed from standard YAML list syntax.
  - Keep existing comma-separated `tools: Read, Grep, Glob` syntax supported.
  
  ## Problem
  `AgentSpecLoader` previously converted the `tools` front matter value with `toString()` before splitting by comma. When `tools` is written as a YAML list, SnakeYAML returns a list, which becomes a 
  string like `[Read, Grep, Glob]`. This causes parsed tool names to include brackets, such as `[Read` and `Glob]`.

  ## Test plan
  - Added regression coverage for YAML list-form `tools`.
  - Ran:
    - `./mvnw -q -pl :spring-ai-alibaba-agent-framework -Dtest=AgentSpecLoaderTest test`
  - Result: currently blocked by unrelated existing compilation errors in skills-related classes, e.g. missing `SkillRegistry#getByPath(...)` and `SkillMetadata#getAllowedTools()`.

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/tools/task/AgentSpecLoader.java
+++ b/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/tools/task/AgentSpecLoader.java
@@ -168,7 +168,7 @@ public final class AgentSpecLoader {
 			return null;
 		}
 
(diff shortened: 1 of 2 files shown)


chensishang:
@yuluo-yx 请帮忙审核一下这个PR，谢谢！

yuluo-yx (approved):
LGTM

PR #4813, opened 2026-07-13
fix(graph): support primitive arrays in deep copy to prevent ClassCastException

### LINKED ISSUE #4812: [BUG]  CompiledGraph.invoke throws ClassCastException when a graph input value is a primitive array (e.g. float[] embedding / byte[] audio)

### Current Behavior

`SerializationUtils.deepCopyValue(Object)` casts every array to `Object[]`. For primitive arrays (`int[]`, `float[]`, `double[]`, `byte[]`, `long[]`, `short[]`, `char[]`, `boolean[]`), this throws a `ClassCastException`, because primitive arrays are **not** subtypes of `Object[]`.

This code path is not wrapped by the surrounding `try/catch`, so the exception propagates out of the public API `CompiledGraph.invoke(...)` and crashes the graph during initial state creation, before any node runs.

The issue is easily triggered because several public APIs naturally return primitive arrays. For example, `EmbeddingModel.embed(...)` returns a `float[]`, while multimodal TTS output is a `byte[]`. Passing either of these values as graph input causes the graph initialization to fail immediately.

The execution flow is:

```text
CompiledGraph.invoke(Map inputs)
    -> CompiledGraph.stateCreate(inputs)
        -> OverAllStateBuilder.withData(inputs)
            -> SerializationUtils.deepCopyMap(inputs)
                -> SerializationUtils.deepCopyValue(value)
                    -> (Object[]) value
                       -> ClassCastException
```

The root cause is that `SerializationUtils.deepCopyValue(Object)` assumes every array can be cast to `Object[]`. This assumption is incorrect because primitive arrays (such as `float[]` and `byte[]`) are not instances of `Object[]`, resulting in an uncaught `ClassCastException` that escapes the public API.

### Expected Behavior

The graph runs normally. Primitive-array input values (e.g., a `float[]` embedding) are deep copied and preserved in the graph state, just like object arrays such as `Integer[]` and `String[]`.

### Steps To Reproduce

1. Build a trivial graph and invoke it with an input map whose value is a primitive array.

```java
KeyStrategyFactory f = () -> {
    Map<String, KeyStrategy> m = new HashMap<>();
    m.put("embedding", (o, n) -> n);
    m.put("result", (o, n) -> n);
    return m;
};

CompiledGraph graph = new StateGraph(f)
        .addNode("echo", node_async(state -> Map.of("result", "ok")))
        .addEdge(START, "echo")
        .addEdge("echo", END)
        .compile();

Map<String, Object> inputs = new HashMap<>();
inputs.put("embedding", new float[] {0.1f, 0.2f, 0.3f}); // e.g. embeddingModel.embed(text)

graph.invoke(inputs);
```

2. A real-world variant produces the same result:

```java
graph.invoke(Map.of("embedding", embeddingModel.embed(text)));
```

For example, DashScope `text-embedding-v2` returns a `float[]` (runtime class `[F`, 1536 dimensions).

3. Observe that `invoke(...)` throws a `ClassCastException` before any graph node is executed.

### Environment

```markdown
Spring AI Alibaba version(s): 1.1.2.2 
Spring AI: 1.1.2
```

### Debug logs

java.lang.ClassCastException: class [F cannot be cast to class [Ljava.lang.Object; ([F and [Ljava.lang.Object; are in module java.base of loader 'bootstrap')
        at com.alibaba.cloud.ai.graph.utils.SerializationUtils.deepCopyValue(SerializationUtils.java:121)
        at com.alibaba.cloud.ai.graph.utils.SerializationUtils.deepCopyMap(SerializationUtils.java:71)
        at com.alibaba.cloud.ai.graph.OverAllStateBuilder.withData(OverAllStateBuilder.java:114)
        at com.alibaba.cloud.ai.graph.CompiledGraph.stateCreate(CompiledGraph.java:704)
        at com.alibaba.cloud.ai.graph.CompiledGraph.invoke(CompiledGraph.java:640)
        at com.alibaba.cloud.ai.graph.CompiledGraph.invoke(CompiledGraph.java:662)

### Anything else?

_No response_

Koiyd:
`SerializationUtils.deepCopyValue(Object)` casts every array to `Object[]`, which throws a `ClassCastException` for primitive arrays (`int[]`, `float[]`, `double[]`, `byte[]`, etc.). The array-handling branch is not guarded by the surrounding `try/catch`, so the exception propagates through the public API:

```text
CompiledGraph.invoke(inputs)
    -> OverAllStateBuilder.withData(inputs)
        -> SerializationUtils.deepCopyMap(inputs)
            -> SerializationUtils.deepCopyValue(value)
```

As a result, the graph crashes during initial state creation before any node is executed.

This is a real and reachable scenario. For example:

- `EmbeddingModel.embed(...)` returns a `float[]`.
- Multimodal TTS output is a `byte[]`.

Passing either value as graph input triggers the crash.

Additionally, the previous implementation silently narrowed reference arrays (e.g. `String[]`) to `Object[]`, losing their runtime component type.

Fixes #4812

### Describe how you did it
Use `java.lang.reflect.Array` to copy arrays while preserving their runtime component type. Primitive arrays (e.g. `int[]`, `float[]`) are copied as their original types, while reference-array elements continue to be deep copied recursively.

```java
if (value.getClass().isArray()) {
    int length = java.lang.reflect.Array.getLength(value);
    Object copyArray = java.lang.reflect.Array.newInstance(value.getClass().getComponentType(), length);
    for (int i = 0; i < length; i++) {
        java.lang.reflect.Array.set(copyArray, i, deepCopyValue(java.lang.reflect.Array.get(value, i)));
    }

    return copyArray;
}
```

### Describe how to verify it
Added `SerializationUtilsPrimitiveArrayTest` (10 tests) covering:

- All primitive array types.
- Empty arrays.
- Deep-copy independence (modifying the copy does not affect the original).
- Runtime component-type preservation (`String[]` remains `String[]`, not `Object[]`).
- Recursive deep copy of reference-array elements.
- The `deepCopyMap` path with a `float[]` value.

Run:

```bash
./mvnw -pl :spring-ai-alibaba-graph-core \
    -Dtest=SerializationUtilsPrimitiveArrayTest test
```

Also verified against the published `1.1.2.2` artifact:

- **Before the fix:** `CompiledGraph.invoke(Map.of("embedding", new float[]{...}))` throws `ClassCastException`.
- **After the fix:** the graph executes successfully and the `float[]` is preserved.

Checkstyle also passes with **0 violations**.

### Special notes for reviews

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/utils/SerializationUtils.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/utils/SerializationUtils.java
@@ -126,13 +126,18 @@ public class SerializationUtils {
 			return copySet;
 		}
 
-		// Handle arrays
-		if (value.getClass().isArray()) {
-			Object[] originalArray = (Object[]) value;
-			Object[] copyArray = new Object[originalArray.length];
-			for (int i = 0; i < originalArray.length; i++) {
-				copyArray[i] = deepCopyValue(originalArray[i]);
-			}
+        // Handle arrays (both object arrays like String[] and primitive arrays like
+        // int[], float[], double[], byte[], ...). Using java.lang.reflect.Array preserves
+        // the original component type and avoids casting a primitive array to Object[],
+        // which throws ClassCastException (e.g. [F cannot be cast to [Ljava.lang.Object;).
+        // Primitive elements are immutable and returned as-is by the recursive call, so a
+        // single element-wise loop is a correct deep copy for both array kinds.
+        if (value.getClass().isArray()) {
+            int length = java.lang.reflect.Array.getLength(value);
+            Object copyArray = java.lang.reflect.Array.newInstance(value.getClass().getComponentType(), length);
+            for (int i = 0; i < length; i++) {
+                java.lang.reflect.Array.set(copyArray, i, deepCopyValue(java.lang.reflect.Array.get(value, i)));
+            }
 			return copyArray;
 		}
 
--- a/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/SerializationUtilsPrimitiveArrayTest.java
+++ b/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/SerializationUtilsPrimitiveArrayTest.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.alibaba.cloud.ai.graph;
+
+import com.alibaba.cloud.ai.graph.utils.SerializationUtils;
+import org.junit.jupiter.api.Test;
+
+import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


guanxuc (approved):
LGTM, thanks!

PR #4816, opened 2026-07-15
fix(graph): deep copy primitive arrays without ClassCastException

### LINKED ISSUE #4812: [BUG]  CompiledGraph.invoke throws ClassCastException when a graph input value is a primitive array (e.g. float[] embedding / byte[] audio)

yyyCode:

`SerializationUtils.deepCopyValue(Object)` casts every array to `Object[]`. Primitive arrays (`int[]`, `float[]`, `double[]`, `byte[]`, `long[]`, `short[]`, `char[]`, `boolean[]`) are **not** subtypes of `Object[]`, so the cast throws a `ClassCastException`. This code path is not wrapped by the surrounding `try/catch`, so the exception escapes the public API `CompiledGraph.invoke(...)` and crashes the graph during initial state creation, before any node runs.

This is easy to hit because several public APIs return primitive arrays — e.g. `EmbeddingModel.embed(...)` returns a `float[]` and multimodal TTS output is a `byte[]`. Passing either as graph input fails immediately.

Fixes #4812

### Describe how you did it

Handle arrays via `java.lang.reflect.Array` in `deepCopyValue`:

- **Primitive arrays**: allocate an array of the same component type and `System.arraycopy` the elements (primitive elements are immutable, so a shallow element copy is a correct deep copy).
- **Object arrays**: allocate an array of the *original* component type (instead of widening to `Object[]`) and recursively `deepCopyValue` each element, preserving the declared array type.

### Describe how to verify it

Added `SerializationUtilsArrayTest` covering:

- `deepCopyValue` on `float[]`, `byte[]`, `int[]` — no exception, values equal, new instance returned
- object arrays preserve their component type (`Integer[]` stays `Integer[]`, not `Object[]`)
- `deepCopyMap` with a primitive-array value
- `CompiledGraph.invoke(...)` with a `float[]` graph input runs end to end and preserves the array in the final state

```shell
./mvnw -pl :spring-ai-alibaba-graph-core -Dtest=SerializationUtilsArrayTest test
```

All 6 new tests pass, and the existing `SerializationUtilsCustomMapTypeTest` still passes.

### Special notes for reviews

The object-array branch now preserves the concrete component type, which is a small behavior improvement over the previous always-`Object[]` result.

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/utils/SerializationUtils.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/utils/SerializationUtils.java
@@ -128,10 +128,21 @@ public class SerializationUtils {
 
 		// Handle arrays
 		if (value.getClass().isArray()) {
-			Object[] originalArray = (Object[]) value;
-			Object[] copyArray = new Object[originalArray.length];
-			for (int i = 0; i < originalArray.length; i++) {
-				copyArray[i] = deepCopyValue(originalArray[i]);
+			Class<?> componentType = value.getClass().getComponentType();
+			int length = java.lang.reflect.Array.getLength(value);
+			// Primitive arrays (int[], float[], byte[], ...) are not instances of
+			// Object[], so casting them would throw a ClassCastException. Their
+			// elements are immutable, so a shallow element copy is sufficient.
+			if (componentType.isPrimitive()) {
+				Object copyArray = java.lang.reflect.Array.newInstance(componentType, length);
+				System.arraycopy(value, 0, copyArray, 0, length);
+				return copyArray;
+			}
+			// Object arrays: preserve the original component type instead of forcing
+			// Object[], and deep copy each element.
+			Object copyArray = java.lang.reflect.Array.newInstance(componentType, length);
+			for (int i = 0; i < length; i++) {
+				java.lang.reflect.Array.set(copyArray, i, deepCopyValue(java.lang.reflect.Array.get(value, i)));
 			}
 			return copyArray;
 		}
--- a/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/SerializationUtilsArrayTest.java
+++ b/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/SerializationUtilsArrayTest.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright 2024-2026 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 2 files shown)


guanxuc:
Thanks for your contribution. Unfortunately, Since there is already an existing PR that addresses this issue, I have to close this one. Would you mind helping to review #4813 instead?

PR #4864, opened 2026-07-28
fix(graph): support primitive arrays in append APIs

keaipiao:

`AppendStrategy` and `AppenderChannel` treat arrays as appendable collections, but both cast every array to a reference array. Primitive arrays such as `float[]` embeddings or `int[]` state values therefore fail with `ClassCastException` instead of being appended.

This change converts arrays through `java.lang.reflect.Array`, which works for both primitive and reference arrays and preserves the existing boxed-list append semantics.

NONE

### Describe how you did it

- Replace the invalid `Object[]`/`T[]` casts with reflective array iteration.
- Apply the same behavior to both the current `AppendStrategy` API and the legacy/public `AppenderChannel` API.
- Add regression tests covering `float[]` and `int[]` inputs.

### Describe how to verify it

```bash
./mvnw -pl spring-ai-alibaba-graph-core -Dtest=PrimitiveArrayAppendTest test
./mvnw -pl spring-ai-alibaba-graph-core '-Dtest=!DatabaseStorePostgreSqlIntegrationTest' test
./mvnw -pl spring-ai-alibaba-graph-core -DskipTests spotless:check checkstyle:check package
```

The graph-core suite completed with 409 tests, 0 failures, and 0 errors after excluding the Docker-only PostgreSQL integration test.

### Special notes for reviews

No external service or model API is required to reproduce or verify this fix. Reference-array behavior remains unchanged.

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/state/AppenderChannel.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/state/AppenderChannel.java
@@ -18,8 +18,8 @@ package com.alibaba.cloud.ai.graph.state;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Supplier;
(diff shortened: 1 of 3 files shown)


oss-sentinel-ai (commented):
## 🔍 Automated Code Review — PR #4864

ENGINE_ERROR: Code engine failed to produce review

---
<sub>🤖 Generated by github-manager | Review ID: heartbeat-20260808T133146Z-1625892</sub>

PR #4911, opened 2026-08-20
fix(graph): notify lifecycle listener on streaming errors

aravelo7:
### Describe what this PR does / why we need it

Fixes a missing `GraphLifecycleListener.onError` notification when a streaming node returns a `Flux` that terminates with an error.

In the current streaming path, the Reactor error signal is converted to `GraphResponse.error(...)`, so it does not reach the existing `CompletableFuture` failure handling that notifies lifecycle listeners. As a result, the original exception is still propagated to the caller, but the lifecycle listener is not notified.

This causes inconsistent lifecycle behavior between normal asynchronous node failures and streaming `Flux` failures.

### Does this pull request fix one issue?

Fixes #4920

### Describe how you did it

Added the missing lifecycle error notification in the streaming error path before returning `GraphResponse.error(...)`.

The change reuses the existing lifecycle listener mechanism and keeps the current exception propagation behavior unchanged.

Regression coverage was also updated to verify that:

- streaming `Flux` failures trigger `onError` exactly once
- the listener receives the expected node id
- the original streaming exception still propagates to the caller
- existing `CompletableFuture` failure behavior remains unchanged

### Describe how to verify it

Run the targeted tests:

```text
StateGraphTest.testStreamingNodeWithFluxException
StateGraphTest.testStreamingNodeWithNodeException
```

### Special notes for reviews

This change only affects the Reactor error path for streaming results.

The existing `CompletableFuture` failure path and exception propagation behavior remain unchanged.

PR #4925, opened 2026-08-25
fix(graph): preserve streaming node id in error callbacks

aravelo7:

Follow-up to #4911, addressing post-merge review feedback on streaming error node attribution.

Streaming errors from `GraphFlux` and `ParallelGraphFlux` can be reported to lifecycle listeners with the outer execution node ID instead of the actual failing stream node ID.

`transformFluxToGraphResponse` already receives the effective stream `nodeId`, but the error path previously dispatched the lifecycle event through `context.doListeners(ERROR, ...)`, which resolved the node ID from the current execution context.

As a result, failures from child or parallel streams could be attributed to their parent node.

NONE

### Describe how you did it

Added an overload of `doListeners` that accepts an explicit node ID while preserving the existing behavior of the original method.

The streaming error path now dispatches lifecycle errors using the effective stream node ID passed to `transformFluxToGraphResponse`.

Added regression tests covering:

- a failing child `GraphFlux`
- a failing child inside `ParallelGraphFlux`

### Describe how to verify it

Ran the four related `StateGraphTest` streaming error tests successfully.

Also verified:

- Checkstyle passes
- `git diff --check` passes

### Special notes for reviews

This change only affects lifecycle error attribution for streaming `GraphFlux` and `ParallelGraphFlux` failures.

Existing callers of `doListeners(scene, error)` retain their original behavior.

### DIFF
--- a/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/GraphRunnerContext.java
+++ b/spring-ai-alibaba-graph-core/src/main/java/com/alibaba/cloud/ai/graph/GraphRunnerContext.java
@@ -314,23 +314,33 @@ public class GraphRunnerContext {
 	// ================================================================================================================
 
 	public void doListeners(String scene, Exception e) {
+		doListeners(scene, getCurrentNodeId(), e);
+	}
+
+	/**
+	 * Notifies lifecycle listeners using an explicit node ID.
+	 * @param scene the lifecycle scene
+	 * @param nodeId the node associated with the lifecycle event
+	 * @param e the error when the scene is {@link StateGraph#ERROR}
+	 */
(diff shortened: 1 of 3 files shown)


oss-sentinel-ai (approved):
LGTM. Clean, focused fix with solid test coverage.

- The new `doListeners(scene, nodeId, e)` overload is backward compatible: the original two-arg method now delegates with `getCurrentNodeId()`, so existing callers keep their behavior.
- The streaming error path in `NodeExecutor` now reports lifecycle errors with the effective stream node ID, so `GraphFlux`/`ParallelGraphFlux` child failures are no longer attributed to the parent node.
- Two regression tests cover both the single child `GraphFlux` and the failing child inside `ParallelGraphFlux`.

---

yuluo-yx (approved):

PR #4904, opened 2026-08-17
fix(graph): align byte array round-trip assertion with serializer behavior

zeng-bohan:
### Problem

`main` CI has been red since #4846 was merged (Aug 15): `SpringAIJacksonStateSerializerTest.shouldPreserveJacksonByteArraySerialization` fails.

### Root cause

#4846 added `instantiateScalarPayloadArray` in `JacksonDeserializer`, which intentionally reconstructs `byte[]` from Jackson's WRAPPER_ARRAY representation (`[B` + base64 payload) so values like `List<byte[]>` round-trip as the original array type (see the javadoc: "so they round-trip as the original array type instead of degrading to a List of [typeId, payload]").

The accompanying test, however, asserted the degraded form (`List.of("[B", "AQID")`), contradicting the implementation it was supposed to protect. The same PR's `shouldPreserveOrdinaryObjectArrayType` test correctly asserts round-tripping as the original array type, so this assertion was simply written wrong.

### Fix

Align the assertion with the documented serializer behavior: assert the payload round-trips as `byte[]` (mirrors `shouldPreserveOrdinaryObjectArrayType`).

### Verification

`mvn -pl spring-ai-alibaba-graph-core test` passes: 464 tests, 0 failures, 0 errors (before the fix: 1 failure).

PR #4926, opened 2026-08-25
test(graph): pin List<byte[]> element types across serializer round-trip

zeng-bohan:
## What

Add a regression test pinning the round-trip behavior of `List<byte[]>` metadata in `SpringAIJacksonStateSerializer`.

## Why

The serializer restores `byte[]` values from Jackson's WRAPPER_ARRAY representation (`[B` descriptor + base64 payload), and this also holds when primitive arrays are nested inside a collection: a `List<byte[]>` survives a full serialize/deserialize cycle as a `List` whose elements are still `byte[]`, matching the documented behavior.

Existing coverage only pins the single-array payload case (`shouldPreserveJacksonByteArraySerialization`, fixed by #4860). While investigating that fix in #4904, it became clear the nested case had no coverage. This test pins it so a future serialization change that degrades elements into `"[B"` string descriptors fails loudly instead of silently.

No production code is changed.

### DIFF
--- a/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/plain_text/SpringAIJacksonStateSerializerTest.java
+++ b/spring-ai-alibaba-graph-core/src/test/java/com/alibaba/cloud/ai/graph/plain_text/SpringAIJacksonStateSerializerTest.java
@@ -489,6 +489,25 @@ class SpringAIJacksonStateSerializerTest {
 	}
 
 	@Test
+	void shouldPreserveListOfByteArraysRoundTrip() throws Exception {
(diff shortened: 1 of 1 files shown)


oss-sentinel-ai (approved):
LGTM. Clean, well-scoped regression test.

The new `shouldPreserveListOfByteArraysRoundTrip` pins the nested `List<byte[]>` round-trip behavior that previously had no coverage, mirrors the existing `shouldPreserveJacksonByteArraySerialization` style, and touches test code only. Assertions verify both element types and payload contents, so a future regression degrading elements into `"[B"` descriptors will fail loudly.

---

yuluo-yx (approved):

PR #4928, opened 2026-08-26
fix(agent): prevent infinite loop when oldString is empty in EditFile

Franctoryer:
# Fix: Prevent infinite loop when `oldString` is empty in `EditFileTool`

## Problem

When using `EditFileTool`, passing an empty string (`""`) as `oldString` causes an **infinite loop** in the `countOccurrences` method.

### Root Cause

```java
private static int countOccurrences(String content, String search) {
    int count = 0;
    int index = 0;
    while ((index = content.indexOf(search, index)) != -1) {
        count++;
        index += search.length();
    }
    return count;
}
```

When `search = ""`:
- `content.indexOf("", index)` always returns `0`
- `search.length()` = `0`, so `index` never advances
- The loop never terminates

## Solution

1. **Replace custom `countOccurrences` with `StringUtils.countMatches`**  
   - Handles empty `search` string safely and returns `0` without looping

2. **Explicitly handle empty file scenario**  
   - When both file content and `oldString` are empty, write `newString` directly to the file
   - This enables editing empty files by simply writing new content

## Result

- Infinite loop is eliminated
- Empty files can now be edited correctly
- Behavior remains consistent with expected file editing semantics

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/extension/tools/filesystem/EditFileTool.java
+++ b/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/extension/tools/filesystem/EditFileTool.java
@@ -15,6 +15,7 @@
  */
 package com.alibaba.cloud.ai.graph.agent.extension.tools.filesystem;
 
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.ai.chat.model.ToolContext;
 import org.springframework.ai.tool.ToolCallback;
 import org.springframework.ai.tool.function.FunctionToolCallback;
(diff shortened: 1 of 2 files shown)


oss-sentinel-ai:
## PR Review — #4928: fix(agent): prevent infinite loop when oldString is empty in EditFile

### Summary
This PR modifies 2 files (+88 -15) on branch `fix_editfile_tool` targeting `main`.

### Changed Files
- `spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/extension/tools/filesystem/EditFileTool.java`
- `spring-ai-alibaba-agent-framework/src/test/java/com/alibaba/cloud/ai/graph/agent/extension/tools/filesystem/EditFileToolTest.java`

### Note
Automated deep review via code engine was unavailable. This is a structural overview only.
Please verify:
- All changed files are intentional
- Tests cover the modifications
- No credentials or secrets are introduced
- API changes are backward compatible

oss-sentinel-ai (approved):
LGTM. Verified against `main`:

- The infinite loop is real: `countOccurrences(content, "")` matches at every position and advances by `search.length() == 0`, never terminating. `StringUtils.countMatches` returns 0 for an empty search string, which cleanly falls into the existing "String not found" path.
- The new empty-file + empty-`oldString` branch gives a sensible semantic (write to empty file) instead of an error.
- `commons-lang3` is already a declared dependency of `spring-ai-alibaba-agent-framework`, so the import is safe.
- Test coverage is good: empty/empty, non-empty/empty, single replace, and replace-all cases.

---

PR #4942, opened 2026-09-01
fix: use object input schema for ls and glob filesystem tools

### LINKED ISSUE #4937: [BUG] ReactAgent添加filesystemInterceptor，使用deepseek模型报错400

### Current Behavior

大模型接口返回的异常信息为：
{
  "error" : {
    "message" : "Invalid schema for function 'ls': schema must be a JSON Schema of 'type: \"object\"', got 'type: \"string\"'.",
    "type" : "invalid_request_error",
    "param" : null,
    "code" : "invalid_request_error"
  }
}
默认打印的异常信息只有org.springframework.web.reactive.function.client.WebClientResponseException$BadRequest: 400 Bad Request from POST https://api.deepseek.com/chat/completions，得手动agent.stream().doOnError()才能获取到原异常信息

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

```markdown
Spring AI Alibaba version(s):
```

### Debug logs

_No response_

### Anything else?

_No response_

CryoThrust:
Closes #4937

## Summary

- Wrap `ListFilesTool` and `GlobTool` single string parameters in request classes (`ListFilesRequest`, `GlobRequest`) with `@JsonProperty` annotations
- Change `inputType` from `String.class` to the request class so the generated JSON schema is `type: object` with a required string property
- This matches the pattern used by `ReadFileTool`, `WriteFileTool`, `EditFileTool`, and `GrepTool`

## Problem

Both tools used `.inputType(String.class)` which produces a top-level JSON schema with `type: "string"`. Providers like DeepSeek reject tool definitions that require `type: "object"`, returning HTTP 400:

```
Invalid schema for function 'ls': schema must be a JSON Schema of 'type: "object"', got 'type: "string"'
```

## Changes

| Tool | Request class | Property |
|------|--------------|----------|
| ls | `ListFilesRequest` | `path` (required) |
| glob | `GlobRequest` | `pattern` (required) |

## Verification

- `ListFilesTool.listFilesContent()` static method is unchanged — callers like `FileSystemTools` are not affected
- The public API of both tools remains backward compatible; only the JSON schema exposed to LLMs changes from string to object

### DIFF
--- a/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/extension/tools/filesystem/GlobTool.java
+++ b/spring-ai-alibaba-agent-framework/src/main/java/com/alibaba/cloud/ai/graph/agent/extension/tools/filesystem/GlobTool.java
@@ -15,9 +15,10 @@
  */
 package com.alibaba.cloud.ai.graph.agent.extension.tools.filesystem;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyDescription;
 import org.springframework.ai.chat.model.ToolContext;
 import org.springframework.ai.tool.ToolCallback;
-import org.springframework.ai.tool.annotation.ToolParam;
 import org.springframework.ai.tool.function.FunctionToolCallback;
 
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


oss-sentinel-ai (approved):
LGTM. Verified against `main`:

- `GlobTool` and `ListFilesTool` were the only two filesystem tools still using `inputType(String.class)`; siblings (`ReadFileTool`, `WriteFileTool`, `EditFileTool`, `GrepTool`) already use dedicated request objects, so this aligns the two stragglers with the established pattern.
- The new `GlobRequest`/`ListFilesRequest` classes follow the same field style as `WriteFileRequest` (public fields + `@JsonProperty(required = true)` + `@JsonPropertyDescription`), producing proper object input schemas as required by most model providers.
- Behavior of the underlying glob/list logic is unchanged.

---

PR #4949, opened 2026-09-07
fix(agent): avoid duplicate instructions in AgentTool

beemines:
### Describe what this PR does / why we need it

When a `ReactAgent` with an instruction runs through `AgentTool`, both the tool executor and the default `InstructionAgentHook` append that instruction. The model receives two copies; with a template such as `Answer about {topic}.`, the first copy remains unrendered because `AgentLlmNode` renders only the last pending instruction.

This change sends the user input through `AgentTool` and leaves instruction injection to the existing hook, matching the subgraph adapter's approach.

### Does this pull request fix one issue?

Fixes #4946

### Describe how you did it

- Remove the manual instruction append and its unused import from `AgentTool`.
- Add `AgentToolInstructionTest`, exercising the public tool callback with a local capturing `ChatModel`. It checks plain instructions, templated instructions supplied with state by a hook, and agents without instructions, each with and without a parent `RunnableConfig`.

### Describe how to verify it

On JDK 17:

```shell
./mvnw -B -pl :spring-ai-alibaba-agent-framework -am -Dtest=AgentToolInstructionTest,AgentToolConfigPropagationTest,AgentToolNodeBuilderTest,AgentToolNodeIntegrationTest,AgentToolNodeSequentialStateIsolationTest,SaaStTemplateRendererTest -Dsurefire.failIfNoSpecifiedTests=false test
```

- Before the fix on `f82da0b`: the new suite ran 6 tests, with 4 assertion failures caused by duplicate/plain or unrendered/template instructions; the 2 no-instruction cases passed.
- After the fix: all 57 selected tests pass, with no errors or skips. Checkstyle reports 0 violations.
- Tested locally on Windows using Maven 3.9.4 and JDK 17.0.17. No model credentials or external model calls are needed for these tests. The full repository test suite was not run.

### Special notes for reviews

The fix follows the single injection point identified in the issue discussion. AI assistance was used for investigation, implementation, and tests; the regression was reproduced locally before applying the fix.

PR #4951, opened 2026-09-07
fix(graph): honor append duplicate policy for scalar updates

### LINKED ISSUE #4950: [BUG] AppendStrategy(false) ignores duplicate policy for scalar updates

- [x] I have searched the existing issues and pull requests.

### Current Behavior

`AppendStrategy(false)` removes duplicates when a node returns a list, but ignores the flag when the node returns a single value. The same logical update therefore produces different state depending on whether it is wrapped in a list.

### Expected Behavior

When appending to an existing list with `allowDuplicate=false`, a scalar update should have the same duplicate policy and encounter order as an equivalent singleton-list update, without mutating the previous state. The default duplicate-allowing behavior should remain unchanged.

### Steps To Reproduce

Using `com.alibaba.cloud.ai.graph.state.strategy.AppendStrategy` and `java.util.List`:

```java
var strategy = new AppendStrategy(false);
var previous = List.of("first", "second");

strategy.apply(previous, "second");         // [first, second, second]
strategy.apply(previous, List.of("second")); // [first, second]
```

This also reproduces through `StateGraph`: register `new AppendStrategy(false)` for a `results` key, connect two nodes that each return `Map.of("results", "answer")`, and invoke the graph. The final state is `[answer, answer]` instead of `[answer]`.

### Environment

- Current `main`: `f82da0b50f35744c13968191be2b1cd2452ef550` (POM version 1.1.2.2)
- Windows, JDK 17.0.17, Maven 3.9.4
- No database or external model calls required

### Debug logs

A local regression suite has 9 cases. Before the fix, 4 assertions fail, including:

```text
graphShouldHonorDuplicatePolicyForScalarNodeUpdates: expected <[answer]> but was <[answer, answer]>
scalarAndSingletonListShouldApplyTheSameDuplicatePolicy: expected <[first, second]> but was <[first, second, second]>
```

### Anything else?

The scalar branch currently returns the appended list without consulting `allowDuplicate`. The list branch already applies `distinct()`. I have reproduced the issue with tests and am preparing a focused fix. This report concerns scalar updates to an existing list, not the separate subgraph delta-merging or primitive-array issues.

beemines:

`AppendStrategy(false)` deduplicates list updates but unconditionally appends scalar updates. Two graph nodes returning the same scalar therefore leave duplicate results even when the key explicitly disallows duplicates. Wrapping the same update in a singleton list already produces the expected result.

This makes scalar appends to an existing list honor the configured duplicate policy, using the same equality and encounter-order semantics as list appends.

Fixes #4950

### Describe how you did it

- Apply `distinct()` to the copied scalar-append result only when `allowDuplicate` is false.
- Add 9 regression cases covering scalar/list equivalence, existing duplicates, immutable and mutable previous lists, optional-wrapped state, default duplicate-allowing behavior, and a compiled graph with two scalar-producing nodes.

### Describe how to verify it

Focused regression:

```shell
./mvnw -B -pl :spring-ai-alibaba-graph-core -am -Dtest=AppendStrategyDeduplicationTest test
```

Broader graph-core regression:

```shell
./mvnw -B -pl :spring-ai-alibaba-graph-core -am '-Dtest=*,!DatabaseStorePostgreSqlIntegrationTest' -DCI=false test
```

On Windows with JDK 17.0.17 and Maven 3.9.4:

- Before the fix on `f82da0b`: 9 new test cases, 4 assertion failures, 0 errors. The compiled graph returned `[answer, answer]` instead of `[answer]` when duplicates were disabled.
- After the fix: 475 cases reported, 402 passed and 73 skipped, 0 failures/errors. All 9 new cases passed. `DatabaseStorePostgreSqlIntegrationTest` was explicitly excluded because it requires Docker; external model credentials were unset for the test process.
- Checkstyle: 0 violations. `git diff --check`: clean.

### Special notes for reviews

Default duplicate-allowing behavior and previous-state immutability are preserved. The change is limited to scalar appends to an existing list; it does not alter replacement/removal handling or primitive-array support. It is independent of #4949 and based directly on upstream `main`.

AI assistance was used for investigation, implementation, and tests; the failing behavior was reproduced locally before applying the fix.


oss-sentinel-ai (commented):
### Summary
This PR modifies 2 file(s) (+110/-1) on branch `fix/append-strategy-scalar-dedup` → `main`.

No issues found. Looks good!

---

PR #4953, opened 2026-09-07
fix(agent): prevent expired parallel tools from starting after timeout

### LINKED ISSUE #4952: [BUG] Timed-out parallel tools can execute after waiting for a permit

- [x] I have searched the existing issues and pull requests.

### Current Behavior

With parallel tool execution and a concurrency limit, a tool waiting in `Semaphore.acquire()` can outlive its outer `CompletableFuture.orTimeout()`:

1. The agent receives timeout responses for both tools.
2. Releasing a permit afterward still lets the waiting tool invoke its callback, despite its response already being finalized as a timeout.
3. If the running tool never releases its permit, the waiter continues occupying an executor worker after the agent call returns.

For tools that perform writes or call external services, the late invocation can cause side effects after the caller has already received a failure. This was reproduced with an invocation counter, not with real external writes.

### Expected Behavior

Waiting for a parallel execution permit should be bounded by the tool's remaining timeout budget, including executor queueing time. A tool whose budget expired while waiting should not begin execution when a permit later becomes available. Waiting workers should become reusable even if an already-running synchronous tool remains blocked.

### Steps To Reproduce

Use an `AgentToolNode` configured as follows:

```java
AgentToolNode.builder()
    .agentName("test-agent")
    .toolCallbacks(List.of(firstTool, secondTool))
    .parallelToolExecution(true)
    .maxParallelTools(1)
    .toolExecutionTimeout(Duration.ofSeconds(2))
    .toolExecutionExceptionProcessor(
        DefaultToolExecutionExceptionProcessor.builder().alwaysThrow(false).build())
    .build();
```

Provide a two-worker executor through `RunnableConfig.addParallelNodeExecutor("_AGENT_TOOL_", executor)` and an assistant message requesting both tools.

- `firstTool` signals a `CountDownLatch` on entry, then waits on another latch before returning.
- `secondTool` increments an `AtomicInteger` when invoked.
- Arrange for the first tool to acquire the sole permit, and observe the second worker waiting in the semaphore.
- Wait for `node.apply(...)` to return its two timeout responses.
- Release the first tool, wait for the executor tasks to finish, and inspect the second tool's invocation count.

Actual: **1**. Expected: **0**.

In a separate test, keep the first tool blocked after the timeout responses and submit a trivial probe task to the same executor. The probe cannot complete because both workers remain occupied.

### Environment

- Current `main`: `f82da0b50f35744c13968191be2b1cd2452ef550` (POM 1.1.2.2)
- Windows, JDK 17.0.17, Maven 3.9.4
- Local fake tools only; no model API or external service needed

### Debug logs

```text
Tests run: 3, Failures: 2, Errors: 0, Skipped: 0
Late invocation: Expecting AtomicInteger(1) to have value: 0
Executor reuse: CompletableFuture[Incomplete] did not complete within 2S
```

The control case, releasing the permit before timeout, passes.

### Anything else?

`orTimeout()` finalizes the future but does not stop the worker already waiting inside `Semaphore.acquire()`. The worker does not check expiry after acquiring the permit. I have a focused fix using timed acquisition and an admission-time expiry check, with the regression tests above and 142 related tests passing locally.

This report concerns tasks still waiting for admission; it does not propose forcefully interrupting synchronous callbacks that were already running.

beemines:

In parallel tool execution, a worker waiting in `Semaphore.acquire()` survives the outer future's timeout. After the agent has returned timeout responses, releasing the permit can still invoke that tool. If the running tool remains blocked, the expired waiter also continues occupying an executor worker.

This bounds admission by the tool's remaining timeout budget, preventing work that expired while waiting from starting later and allowing waiting workers to be reused.

Fixes #4952

### Describe how you did it

- Capture each tool's submission time using the monotonic clock.
- Replace unbounded permit acquisition with `tryAcquire` using the remaining budget, including executor queueing time.
- Recheck the finalized response and deadline after acquiring a permit, before entering the interceptor/tool execution path.
- Reuse the existing timeout handler for error responses, update-map disposal, and cancellation propagation; acquired permits are released through the existing `finally` block.

`AgentToolNodeParallelAdmissionTest` uses two workers, one permit, and latches to hold the first tool. It observes the second worker waiting in the semaphore, then verifies:

1. A timed-out waiter is never invoked after the first tool releases its permit.
2. Its worker becomes reusable while the first tool is still blocked.
3. Releasing the permit before timeout still executes both tools once, with responses in request order.

### Describe how to verify it

```shell
./mvnw -B -pl :spring-ai-alibaba-agent-framework -am '-Dtest=AgentToolNodeParallelAdmissionTest' test
./mvnw -B -pl :spring-ai-alibaba-agent-framework -am '-Dtest=AgentToolNode*Test,ToolStateCollectorTest,DefaultCancellationTokenTest,AsyncToolCallbackAdapterTest,CancellableAsyncToolCallbackTest,AgentToolConfigPropagationTest' test
```

Windows, JDK 17.0.17, Maven 3.9.4:

- Before the fix on `f82da0b`: 3 regression cases, 2 assertion failures, 0 errors. The expired tool's invocation count was 1 instead of 0, and the executor probe could not complete. The pre-timeout control case passed.
- After the fix: all 142 selected tests pass, with 0 failures, errors, or skips. These include existing parallel execution, starvation, cancellation, async-tool, and state-collection tests.
- Checkstyle: 0 violations. `git diff --check`: clean.

### Special notes for reviews

This change covers tasks waiting for admission. It does not forcibly stop synchronous callbacks already executing, or introduce a new executor. Tests use fake tools and no external services/model credentials. The full multi-module test suite was not run.

The branch is based directly on upstream `main` and is independent of #4949 and #4951. AI assistance was used for investigation, implementation, and tests; both failures were reproduced locally before applying the fix.


oss-sentinel-ai (commented):
### Summary
This PR modifies 2 file(s) (+200/-2) on branch `fix/parallel-tool-timeout-admission` → `main`.

No issues found. Looks good!

---

PR #4976, opened 2026-09-16
feat(admin): support model type selection and local docker/frontend f…

guofengxu:
…ixes

Allow configuring llm/embedding/rerank when adding models, and fix local compose, nginx, and frontend image build issues.


guofengxu:
update compose.xml

PR #4558, opened 2026-04-16
fix(SummarizationHook): preserve original SystemMessage during summarization

wade-microblueworld:
### Root Cause

When compressing conversation history, `SummarizationHook` was including **all** messages before the cutoff in the summarization list, including any `SystemMessage` instances. The original system prompt was sent to the LLM for summarization, and its content was replaced with the summarized version, losing critical instruction content.

### Fix

Preserve all `SystemMessage` instances before the cutoff index and prepend them at the top of the new message list, before the summary message. This ensures the original system instructions remain intact.

Fixes #4048

PR #4978, opened 2026-09-17
fix(agent): include tool calls and results in summarization input

luguanxing:

`SummarizationHook.createSummary()` currently serializes each message using only `Message.getText()`.

For tool interactions, this omits two important parts of the conversation:

- `AssistantMessage.getToolCalls()`: tool names, call IDs, and arguments.
- `ToolResponseMessage.getResponses()`: tool results, including retrieved facts and source URLs. `ToolResponseMessage.getText()` is empty even when its responses contain data.

Consequently, the summarization model can receive only `Assistant:` and `Tool:` lines for an otherwise successful retrieval. Once the older messages are replaced by the summary, that tool evidence is no longer available in the conversation. This is an input-serialization issue, not a model-quality issue.

NONE. The offline regression tests in this PR reproduce the bug.

### Describe how you did it

- Retain the existing role/text representation, treating null text as empty.
- Append every assistant tool call with its ID, name, and arguments.
- Append every tool response with its ID, name, and response data.
- Keep IDs on both sides so multiple calls to the same tool remain distinguishable, including responses returned in a different order.
- Leave history cutoff, tool-call pairing, recent-message retention, summary prompts, and error handling unchanged.

Example of information now included in the summary input:

```text
Assistant:
Tool call [id=call-3, name=search]: {"query":"source"}
Tool:
Tool result [id=call-3, name=search]: {"fact":"SENTINEL_EVIDENCE","url":"https://example.org/source"}
```

### Describe how to verify it

```bash
mvn -B -ntp -pl spring-ai-alibaba-agent-framework -am \
  -Dtest=SummarizationHookTests \
  -Dsurefire.failIfNoSpecifiedTests=false clean package
```

The tests use a mocked `ChatModel` and capture the actual summary prompt. No API keys, network searches, or live LLM calls are needed to execute them.

Coverage includes ordinary text and roles, null/empty assistant text, mixed text and tool calls, JSON tool results with source URLs, multiple calls/responses, preserved-message exclusion, and the below-threshold path.

- Before the fix: 8 tests run, 5 failures, 0 errors, 0 skipped.
- After the fix: 8 tests run, 0 failures, 0 errors, 0 skipped.

### Special notes for reviews

This is deliberately limited to serialization of tool interactions into the summarization input. It does not guarantee that an LLM summary will retain every fact, and it does not change the separate summary-error handling behavior.

The SystemMessage preservation change in #4558 addresses a different issue; this patch does not modify `beforeModel()` or the cutoff logic.

The selected offline regression suite was run locally with Temurin JDK 21 (compilation target: Java 17) and Maven 3.9.11. The full project integration-test suite was not run.


luguanxing:
Done: also ported this fix to https://github.com/agentic-spring-ai/agentic-spring-ai/pull/92, following the invitation in #4979. Verified on the target repository: the 8 regression tests reproduce 5 failures before the fix and all pass afterward; clean package and Checkstyle pass. Keeping this original PR open for review here.

